PR #2, opened 2016-05-10
openssl: add PPC args

mistydemeo:
If you'd rather I push these directly instead of opening PRs just let me know!


xu-cheng:
Yes, feel free to push commit directly.

PR #3, opened 2016-05-10
Suggest building on Tiger on PPC

mistydemeo:
Since Tigerbrew supports using Tiger, suggest building on Tiger instead of Leopard there.

I still need to do some testing on Tiger to make sure all is working smoothly.

### DIFF
--- a/Abstract/portable-formula.rb
+++ b/Abstract/portable-formula.rb
@@ -1,10 +1,11 @@
 module PortableFormulaInstall
   def install
-    if OS.mac? && OS::Mac.version > :leopard
+    tag = Hardware::CPU.ppc? ? :tiger : :leopard
+    if OS.mac? && OS::Mac.version > tag
       opoo <<-EOS.undent
         You are building portable formula on #{OS::Mac.version}.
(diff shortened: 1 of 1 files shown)


UniqMartin:
Would it be practical to build on a newer OS X to benefit from a more modern toolchain (as long as it still supports the targeted minium OS X version) and just set the deployment target to a lower value than the current machine we're running on? That's not standard Homebrew behavior, but I figure this could make sense for these special formulae.

Would that even work or are there practical limitations to this approach? (I only started using Macs with 10.5 and it was only much later when I started paying closer attention to the toolchain and how software development usually works on OS X, so what I know now might not be applicable to legacy OS X.)

xu-cheng:
> Would it be practical to build on a newer OS X to benefit from a more modern toolchain (as long as it still supports the targeted minium OS X version) and just set the deployment target to a lower value than the current machine we're running on? That's not standard Homebrew behavior, but I figure this could make sense for these special formulae.

I think this is theoretical possible. However, I don't know how to make it work in practice.

PR #4, opened 2016-08-10
Update documentation, license, scripts.

MikeMcQuaid:
Get the final things ready for the final release in https://github.com/Homebrew/brew/pull/513.

CC @xu-cheng and @Homebrew/maintainers for review.


xu-cheng:
:+1:

Technically this means we will need homebrew/brew instead of tigerbrew to build vendor ruby. Also we could use `brew test-bot --skip-relocation` directly.

As I have tested that `homebrew/brew` can be indeed working on os x 10.5 if we remove the os x version check in `brew.rb`. So should we just remove that os x version check?

MikeMcQuaid:
> As I have tested that homebrew/brew can be indeed working on os x 10.5 if we remove the os x version check in brew.rb. So should we just remove that os x version check?

Yes, seems like a good idea, I was thinking the same.

MikeMcQuaid:
@xu-cheng Done in https://github.com/Homebrew/brew/pull/734.

PR #5, opened 2016-08-18
using test-bot directly to build

xu-cheng:
test-bot will also handle bottle upload as well as revision bump.


MikeMcQuaid:
I don't think this is a 1:1 functionality mapping. If you'd like to move this logic into `test-bot` I think I'd be 👍 on that but we're not there yet and I'm not sure I personally see the advantage.

xu-cheng:
> I don't think this is a 1:1 functionality mapping.

What function you are referring to?

>  I'm not sure I personally see the advantage.

It enables us, say, testing PR in this repo. And more importantly to handle bottle upload as well bottle block upload.

cmd/brew-portable-package.rb
```diff
@@ -1,28 +0,0 @@
...
-if RUBY_VERSION.split(".").first.to_i < 2
-  safe_system "brew", "install", "ruby"
-  ENV["HOMEBREW_RUBY_PATH"] = "#{HOMEBREW_PREFIX}/bin/ruby"
-end
-
-ENV["HOMEBREW_PREFER_64_BIT"] = "1"
-ENV["HOMEBREW_DEVELOPER"] = "1"
-
-ARGV.named.each do |name|
-  name = "portable-#{name}" unless name.start_with? "portable-"
```
MikeMcQuaid: This is not done by the test-bot and should be before this is merged.

cmd/brew-portable-package.rb
```diff
@@ -1,28 +0,0 @@
...
-# Usage: brew portable-package <formula> [formula...]
-
-odie "Need to specify at least one portable formula!" if ARGV.empty?
-
-if RUBY_VERSION.split(".").first.to_i < 2
-  safe_system "brew", "install", "ruby"
-  ENV["HOMEBREW_RUBY_PATH"] = "#{HOMEBREW_PREFIX}/bin/ruby"
-end
-
-ENV["HOMEBREW_PREFER_64_BIT"] = "1"
```
MikeMcQuaid: This is not done by the test-bot and should be before this is merged.

cmd/brew-portable-package.rb
```diff
@@ -1,28 +0,0 @@
-# Build a portable Homebrew utility bottle.
-#
-# Usage: brew portable-package <formula> [formula...]
-
-odie "Need to specify at least one portable formula!" if ARGV.empty?
-
-if RUBY_VERSION.split(".").first.to_i < 2
-  safe_system "brew", "install", "ruby"
```
MikeMcQuaid: This is not done by the test-bot and should be before this is merged.

cmd/brew-portable-package.rb
```diff
@@ -1,28 +0,0 @@
...
-end
-
-ENV["HOMEBREW_PREFER_64_BIT"] = "1"
-ENV["HOMEBREW_DEVELOPER"] = "1"
-
-ARGV.named.each do |name|
-  name = "portable-#{name}" unless name.start_with? "portable-"
-  f = Formula[name]
-  safe_system "brew", "install", "--build-bottle", name
-  safe_system "brew uninstall --force $(brew deps --include-build #{name})"
```
MikeMcQuaid: This is not done by the test-bot and should be before this is merged.

cmd/brew-portable-package.rb
```diff
@@ -1,28 +0,0 @@
...
-  safe_system "brew", "test", name
-  puts "Library linkage:"
```
MikeMcQuaid: This is not output by default by the test-bot and should be before this is merged.

cmd/brew-portable-package.rb
xu-cheng: This is unnecessary. As homebrew/brew already starts to vendor ruby 2.0 by default.

cmd/brew-portable-package.rb
MikeMcQuaid: Ideally `brew linkage` would use `ldd` on Linux.

MikeMcQuaid:
Have commented on the stuff that `test-bot` currently doesn't seem to do. Ideally a `--portable` could be passed which would set all these values.

cmd/brew-portable-package.rb
MikeMcQuaid: It may be necessary on Tigerbrew in future but I guess that will depend on 2.0 at some point anyway so 👍

cmd/brew-portable-package.rb
xu-cheng: This is already documented by README.md as the beginning part of invoking command.

I also don't think this should be added in `test-bot`. As in fact, you can build 32bit only portable files if it's requested.

cmd/brew-portable-package.rb
xu-cheng: Why this is needed? At the same time, `test-bot` will only check formula in this tap. If we set up CI, test-bot will only test formula changed by PR.

cmd/brew-portable-package.rb
xu-cheng: This is included in test-bot. We test formula without build deps since day one.

cmd/brew-portable-package.rb
MikeMcQuaid: code > documentation.

cmd/brew-portable-package.rb
xu-cheng: FWIW, this removed script cannot be run on Tigerbrew yet either. Because `--skip-reloation` flag in `brew bottle` is not supported by it.

cmd/brew-portable-package.rb
xu-cheng: This is covered by `brew test` for OS X. I'm working on another PR to add linux support.

cmd/brew-portable-package.rb
MikeMcQuaid: Definitely not since day one 😉. To be exact: it only uninstalls `unchanged_build_dependencies`, not all of them (which is what we need in this case).

cmd/brew-portable-package.rb
xu-cheng: I mean it is the code and document at the same time:
```
HOMEBREW_PREFER_64_BIT=1 brew test-bot --skip-relocation --tap=homebrew/portable <formula>
```

cmd/brew-portable-package.rb
MikeMcQuaid: We should also output this information for manual inspection, too.

cmd/brew-portable-package.rb
MikeMcQuaid: I'm saying `brew test-bot --portable ruby` should be sufficient if you want to remove this command.

cmd/brew-portable-package.rb
xu-cheng: Do we need it if `brew test` already checks no unwanted linkage invoked?

cmd/brew-portable-package.rb
xu-cheng: I still think this should be tagged as `--portable` flag. How about `brew test-bot --prefer-64bit`.

MikeMcQuaid: When is this required or not?

xu-cheng:
I think I will create a new PR for this.

xu-cheng:
@MikeMcQuaid Sorry to bother you. However, I think I don't have the necessary access to this repo. As result, I cannot delete the outdated branch. Would you mind to give me the admin access to this repo? Thanks.

MikeMcQuaid:
@xu-cheng No bother at all, my fault. You were in the homebrew-portable team but it didn't have any repos added to it, doh! Fixed now.

xu-cheng:
Many thanks.

PR #7, opened 2017-01-25
docker: don't run brew as sudo

### LINKED ISSUE #6: Can't build on Linux
Trying to build the docker image fails because it tries to run Linuxbrew as root, which Linuxbrew no longer allows.

alyssais:
Since Linuxbrew can no longer be run with root privileges, we need to change to the linuxbrew user before running any brew commands.

Fixes #6.


DoomHammer:
Great!

PR #8, opened 2017-01-26
portable-readline: fix patches

alyssais:
The gist this used to use has been taken down, so get the patches directly from GNU.

After this PR is merged, `portable-ruby` should be able to build again on macOS (but not linux).

### DIFF
--- a/Formula/portable-readline.rb
+++ b/Formula/portable-readline.rb
@@ -8,9 +8,37 @@ class PortableReadline < PortableFormula
   version "6.3.8"
   sha256 "56ba6071b9462f980c5a72ab0023893b65ba6debb4eeb475d7a563dc65cafd43"
 
-  patch do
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
🆒 although CC @ilovezfs for help on whether it's possible to use your cool patch tarball thing for this

sjackman:
> After this PR is merged, portable-ruby should be able to build again on macOS (but not linux).

What's holding back Linux? I'll help out if I can.

alyssais:
@sjackman just https://github.com/Homebrew/homebrew-portable/pull/11 (I want to do a little more testing) ;)

mistydemeo:
Are we ready to include this?

alyssais:
Is ready from my point of view unless @ilovezfs wants to do something with the patches (I'm guessing not though since it's been some time).

PR #9, opened 2017-01-26
portable-package: fix incorrect variable name

alyssais:
This stops portable-package from working at all on Linux.

### DIFF
--- a/cmd/brew-portable-package.rb
+++ b/cmd/brew-portable-package.rb
@@ -20,7 +20,7 @@ ARGV.named.each do |name|
   safe_system "brew", "test", name
   puts "Library linkage:"
   if OS.linux?
-    puts Utils.open_read "ldd", "#{f.bin}/#{arg.gsub(/^portable-/, "")}"
+    puts Utils.open_read "ldd", "#{f.bin}/#{name.gsub(/^portable-/, "")}"
   else
     puts Utils.popen_read "brew", "linkage", name
   end


MikeMcQuaid:
Heh. Suggestion: Travis CI for this repo?

alyssais:
> Heh. Suggestion: Travis CI for this repo?

Only really useful when we have it in a working state ;)

DoomHammer:
Not quite. Having it broken and aiming for a fix is fine too.

MikeMcQuaid:
@alyssais Also: removing the things you aren't personally using in here.

alyssais:
@MikeMcQuaid I've actually used everything in here _this week_! I've been doing some work with AWS Lambda, for which an easy way to compile portable binaries is _really_ useful. I even made a [new formula](https://github.com/alyssais/homebrew-portable/blob/6714bc1781baab3ef6ec1045a062221f0da9ef20/Formula/portable-jq.rb) for another utility I needed, though I'm going to keep that on my fork for now since I get the impression you wouldn't be so keen on the scope of this tap _expanding_!

MikeMcQuaid:
@alyssais 🆒. CI would be great when you get a chance.

DoomHammer:
Formula for `jq` is great! @sjackman is experimenting with AWS Lambda as well. Shaun, do you think more utils might be helpful in any way?

@alyssais Great job!

MikeMcQuaid:
@alyssais Might be worth pulling everything here except Ruby into your tap.

alyssais:
@MikeMcQuaid is Ruby the only thing `brew` uses?

@mistydemeo mentioned to me that Tigerbrew uses some other things in here, so if we can find out what those are I'd like to leave those here as well.

mistydemeo:
Tigerbrew uses Ruby (and its dependencies), and Curl.

alyssais:
So that leaves only `portable-git` and `portable-expat`.

mistydemeo:
`portable-git` (and by extension expat) would be useful, but not crucial, to Tigerbrew.

alyssais:
In that case, I'd rather everything that's currently here continues to live under Homebrew. I don't think it would be a good idea for me to be singularly responsible for maintaining stuff that's important to a large user base in a context I have no experience in.

MikeMcQuaid:
@alyssais That's fine, thanks.

sjackman:
@alyssais Hi, Alyssa. I'm also running Linuxbrew in Lambda. Here's the repo: https://github.com/Linuxbrew/linuxbrew-lambda
I'm using it to receive a webhook notification form CircleCI, download the artifacts (bottles), and upload them to BinTray and the SHA256 to GitHub.

PR #10, opened 2017-01-26
portable-package: fix another Linux typo

alyssais:
🙃

### DIFF
--- a/cmd/brew-portable-package.rb
+++ b/cmd/brew-portable-package.rb
@@ -20,7 +20,7 @@ ARGV.named.each do |name|
   safe_system "brew", "test", name
   puts "Library linkage:"
   if OS.linux?
-    puts Utils.open_read "ldd", "#{f.bin}/#{name.gsub(/^portable-/, "")}"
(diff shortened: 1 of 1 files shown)


cmd/brew-portable-package.rb
```diff
@@ -20,7 +20,7 @@
   safe_system "brew", "test", name
   puts "Library linkage:"
   if OS.linux?
-    puts Utils.open_read "ldd", "#{f.bin}/#{name.gsub(/^portable-/, "")}"
+    puts Utils.popen_read "ldd", "#{f.bin}/#{name.gsub(/^portable-/, "")}"
```
sjackman: Linuxbrew has a working `linkage` command now.

MikeMcQuaid: This doesn't assume Linuxbrew, though, but just Homebrew/brew's Linux support, no?. I'd rather avoid relying on commands like that until they are in upstream Homebrew.

sjackman: Ah, right. Good point. I have a conference in week. After I'm back from that I'll send some patches upstream.

PR #11, opened 2017-01-27
Fix portable-ruby (and portable-openssl) on Linux

alyssais:
671cb5c4e85d2d18ea40e70336804fa7d3bc2283 broke `portable-openssl` (and therefore `portable-ruby`) builds on Linux by removing the `zlib-devel` yum package from the Docker image. This package was supposed to be used just for the bootstrap Ruby compilation using ruby-build, but the `portable-openssl` package also ended up using it (presumably accidentally). This meant that portable-openssl had a dynamic link to zlib, which caused the portable-ruby build to fail.

Here's what I did to fix this:

1. Add a simple test to `portable-openssl` to check that zlib support was present and working correctly.
2. Make `portable-openssl` depend on Linuxbrew's `zlib` at build time (on Linux only).
3. Add `zlib`'s prefix to the linker for OpenSSL as a configure flag (on Linux only). (It ignores `LDFLAGS` according to https://github.com/Linuxbrew/legacy-linuxbrew/issues/897, and portable formulae [reset](https://github.com/Homebrew/homebrew-portable/blob/43b2f06293e56dcfd0ada976f1006c88b1b20251/Abstract/portable-formula.rb#L12-L19) these environment variables anyway to avoid accidental dynamic dependencies on Linuxbrew formulae. In this case, `zlib` will be linked statically, so it's okay.)
4. Use OpenSSL's `zlib` configure option instead of `zlib-dynamic` (on Linux only) so OpenSSL links `zlib` statically instead of dynamically.
5. Make `portable-ruby` depend on `zlib` at build time (on Linux only), because now that `portable-openssl` is statically linked against `zlib` Ruby won't pick up `zlib` through `openssl`.
6. Force `portable-ruby` into linking with `zlib` through `LDFLAGS`. (Ruby doesn't usually try to do this because it expects to get it through OpenSSL. I found a bug report on this from 2006!)

Note: I used `unless OS.mac?` for all of these changes instead of `if OS.linux?` because it seems to be like macOS is the special case for being guaranteed to come with zlib. I don't know what the general Linuxbrew style is here (is `if` always preferred?) so I'm happy to change it if such a thing exists.

### DIFF
--- a/Formula/portable-openssl.rb
+++ b/Formula/portable-openssl.rb
@@ -9,6 +9,7 @@ class PortableOpenssl < PortableFormula
   sha256 "1d4007e53aad94a5b2002fe045ee7bb0b3d98f1a47f8b2bc851dcd1c74332919"
 
   depends_on "makedepend" => :build
+  depends_on "zlib" => :build unless OS.mac?
 
   resource "cacert" do
     # homepage "http://curl.haxx.se/docs/caextract.html", "https://github.com/bagder/ca-bundle"
@@ -41,7 +42,8 @@ class PortableOpenssl < PortableFormula
       --prefix=#{prefix}
       --openssldir=#{openssldir}
       no-ssl2
-      zlib-dynamic
+      #{"-L#{Formula["zlib"].opt_prefix/"lib"}" unless OS.mac?}
+      #{OS.mac? ? "zlib-dynamic" : "zlib"}
       no-shared
       enable-cms
     ]
@@ -105,5 +107,10 @@ class PortableOpenssl < PortableFormula
     end
 
     openssldir.install resource("cacert").files("ca-bundle.crt" => "cert.pem")
+  end
+
+  test do
+    input = "x\x9CK\xCB\xCF\a\x00\x02\x82\x01E"
+    assert_equal "foo", pipe_output("#{bin}/openssl zlib -d", input)
   end
 end
--- a/Formula/portable-openssl.rb
+++ b/Formula/portable-openssl.rb
@@ -110,7 +110,8 @@ class PortableOpenssl < PortableFormula
   end
 
   test do
+    cp_r Dir["#{prefix}/*"], testpath
     input = "x\x9CK\xCB\xCF\a\x00\x02\x82\x01E"
-    assert_equal "foo", pipe_output("#{bin}/openssl zlib -d", input)
+    assert_equal "foo", pipe_output("#{testpath}/bin/openssl zlib -d", input)
   end
 end
--- a/Formula/portable-openssl.rb
+++ b/Formula/portable-openssl.rb
@@ -9,7 +9,7 @@ class PortableOpenssl < PortableFormula
   sha256 "1d4007e53aad94a5b2002fe045ee7bb0b3d98f1a47f8b2bc851dcd1c74332919"
 
   depends_on "makedepend" => :build
-  depends_on "zlib" => :build unless OS.mac?
+  depends_on "portable-zlib" => :build unless OS.mac?
 
   resource "cacert" do
     # homepage "http://curl.haxx.se/docs/caextract.html", "https://github.com/bagder/ca-bundle"
@@ -42,7 +42,7 @@ class PortableOpenssl < PortableFormula
       --prefix=#{prefix}
       --openssldir=#{openssldir}
       no-ssl2
-      #{"-L#{Formula["zlib"].opt_prefix/"lib"}" unless OS.mac?}
+      #{"-L#{Formula["portable-zlib"].opt_prefix/"lib"}" unless OS.mac?}
       #{OS.mac? ? "zlib-dynamic" : "zlib"}
       no-shared
       enable-cms
--- a/Formula/portable-ruby.rb
+++ b/Formula/portable-ruby.rb
@@ -14,23 +14,21 @@ class PortableRuby < PortableFormula
 
   depends_on "makedepend" => :build
   depends_on "pkg-config" => :build
-  depends_on "zlib" => :build unless OS.mac?
+  depends_on "portable-zlib" => :build unless OS.mac?
   depends_on "portable-readline" => :build
   depends_on "portable-libyaml" => :build
   depends_on "portable-openssl" => :build
   depends_on "portable-ncurses" => :build if OS.linux?
 
   def install
-    ENV.append "LDFLAGS", "-Wl,-search_paths_first"
-
-    unless OS.mac?
-      ENV.append "LDFLAGS", "-lz -L#{Formula["zlib"].opt_prefix/"lib"}"
-    end
-
+    zlib = Formula["portable-zlib"]
     readline = Formula["portable-readline"]
     libyaml = Formula["portable-libyaml"]
     openssl = Formula["portable-openssl"]
     ncurses = Formula["portable-ncurses"]
+
+    ENV.append "LDFLAGS", "-Wl,-search_paths_first"
+    ENV.append "LDFLAGS", "-lz -L#{zlib.opt_prefix/"lib"}" unless OS.mac?
 
     args = %W[
       --prefix=#{prefix}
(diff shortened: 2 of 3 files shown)


MikeMcQuaid:
Is there not a static system zlib available on Linux? If not, can you not assume zlib is installed on Linux pretty safely?

alyssais:
> Is there not a static system zlib available on Linux? If not, can you not assume zlib is installed on Linux pretty safely?

No, I don't think so. That's why this broke in the first place! There's no point ending up with a nearly-portable-but-not-quite build.

xu-cheng:
Thanks @alyssais for improving this repo, however may I suggest to create a `portable-zlib` formula, which only builds static version zlib. Therefore, it will follow the similar behaviour as rest of formulae like `portable-readline`, and you can make sure no dynamic zlib is accidentally linked.

alyssais:
@xu-cheng happy to try that, but isn't uninstalling zlib and then running the tests enough to verify that it wasn't linked statistically?

xu-cheng:
@alyssais On the one hand, it is more aligned with the rest of formulae in this repo, i.e., all static libs have their own portable formulae. On the other hand, it offers a stronger guarantee that it won't be accidently linked/associated to the dynamic lib as it never exists at the first place. Noted that the test case may not be sufficiently caught problem in some rare edge case.

alyssais:
@xu-cheng makes sense. Thanks. 👍

Formula/portable-openssl.rb
```diff
@@ -41,7 +42,8 @@ def configure_args
       --prefix=#{prefix}
       --openssldir=#{openssldir}
       no-ssl2
-      zlib-dynamic
+      #{"-L#{Formula["zlib"].opt_prefix/"lib"}" unless OS.mac?}
```
xu-cheng: This should be `portable-zlib` right?

Formula/portable-ruby.rb
```diff
@@ -22,6 +23,10 @@ class PortableRuby < PortableFormula
   def install
     ENV.append "LDFLAGS", "-Wl,-search_paths_first"
 
+    unless OS.mac?
+      ENV.append "LDFLAGS", "-lz -L#{Formula["zlib"].opt_prefix/"lib"}"
+    end
```
xu-cheng: `portable-zlib`

alyssais:
Haven't had a chance to test on Linux yet, so please don't merge but let me know if it looks good.

Formula/portable-openssl.rb
```diff
@@ -106,4 +108,9 @@ def install
 
     openssldir.install resource("cacert").files("ca-bundle.crt" => "cert.pem")
   end
+
+  test do
+    input = "x\x9CK\xCB\xCF\a\x00\x02\x82\x01E"
+    assert_equal "foo", pipe_output("#{bin}/openssl zlib -d", input)
+  end
```
xu-cheng: It is custom to run test in following way for formulae in this repo:
```ruby
  cp_r Dir["#{prefix}/*"], testpath
  system testpath/"bin/openssl", ...
```

Formula/portable-zlib.rb
```diff
@@ -0,0 +1,31 @@
...
+  sha256 "c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1"
+
+  # http://zlib.net/zlib_how.html
+  resource "test_artifact" do
+    url "http://zlib.net/zpipe.c"
+    version "20051211"
+    sha256 "68140a82582ede938159630bca0fb13a93b4bf1cb2e85b08943c26242cf8f3a6"
+  end
+
+  def install
```
xu-cheng: May want to add `ENV.universal_binary if build.with? "universal"` to align with the behaviours of other lib formulae in the repo

Formula/portable-openssl.rb
alyssais: Ah, yeah. That's what happens when I don't test on Linux.

MikeMcQuaid:
> Haven't had a chance to test on Linux yet

A Travis CI file calls you 😉. If you want a hand with it: shout.

alyssais:
@MikeMcQuaid I'm working on it :)

Formula/portable-ruby.rb
```diff
@@ -14,19 +14,22 @@ class PortableRuby < PortableFormula
...
-    ENV.append "LDFLAGS", "-Wl,-search_paths_first"
-
+    zlib = Formula["portable-zlib"]
     readline = Formula["portable-readline"]
     libyaml = Formula["portable-libyaml"]
     openssl = Formula["portable-openssl"]
     ncurses = Formula["portable-ncurses"]
 
+    ENV.append "LDFLAGS", "-Wl,-search_paths_first"
+    ENV.append "LDFLAGS", "-lz -L#{zlib.opt_prefix/"lib"}" unless OS.mac?
```
xu-cheng: I'm not sure. But it is seem that you can add zlib lib using `args << "--with-opt-dir=#{paths.join(":")}"` in below, instead of hacking the LDFLAGS. Can you test it? @alyssais

sjackman:
We build Linuxbrew bottles on CircleCI. Would you like to do the same for `Homebrew/portable`?

sjackman:
Ah, I see you're using CentOS 5 in Docker. CircleCI has some support for Docker. https://circleci.com/docs/docker/
I believe they're expanding this support for CircleCI 2.0 now in beta.

alyssais:
@sjackman I actually have Travis working in my fork (using Docker). What made you use Circle for bottles?

sjackman:
TravisCI doesn't support uploading artifacts from fork pull requests (last I checked, many months ago).
CircleCI stores artifacts in their own S3 bucket, and we use a webhook running on AWS Lambda to download them from their and upload them to BinTray. https://github.com/Linuxbrew/linuxbrew-lambda

MikeMcQuaid:
@sjackman Out of interest is there a reason you don't use e.g. CircleCI's S3 bucket for hosting the bottles (behind authentication or something)? Also, if you haven't already, it may be worth requesting to CircleCI that they support Bintray uploading. Would be huge if we could get this be a generic process that can be done for both macOS and Linux taps without needing Lambda.

sjackman:
> Out of interest is there a reason you don't use e.g. CircleCI's S3 bucket for hosting the bottles (behind authentication or something)?

It's easier to maintain if Linuxbrew's infrastructure is similar to Homebrew's. If CircleCI plan to store these artifacts indefinitely without an limit on space or I/O, that's an option. The URL is a bit unwieldy, but it could be stored in the formula. e.g.

https://3631-56942570-gh.circle-artifacts.com/0/tmp/circle-artifacts.0oc1JWR/gperftools-2.5_1.x86_64_linux.bottle.tar.gz

> Also, if you haven't already, it may be worth requesting to CircleCI that they support Bintray uploading.

I did ask, but nothing so far. Trickier is pushing the git commit to BrewTestBot. I submitted a feature request to be able to run an arbitrary script to upload artifacts using credentials that are not exposed to the build of the fork pull request.

alyssais:
I haven't made much progress with this recently because this patch:

```diff
diff --git a/Formula/portable-openssl.rb b/Formula/portable-openssl.rb
index fc37e16883..f3dd177c8f 100644
--- a/Formula/portable-openssl.rb
+++ b/Formula/portable-openssl.rb
@@ -110,7 +110,8 @@ class PortableOpenssl < PortableFormula
   end
 
   test do
+    cp_r Dir["#{prefix}/*"], testpath
     input = "x\x9CK\xCB\xCF\a\x00\x02\x82\x01E"
-    assert_equal "foo", pipe_output("#{bin}/openssl zlib -d", input)
+    assert_equal "foo", pipe_output("#{testpath}/bin/openssl zlib -d", input)
   end
 end
```

somehow breaks `portable-ruby`'s build and I have absolutely no idea why. I'll keep investigating but just wanted to give an update.

sjackman:
Is it able to find its libraries at run time in `testpath/"lib"`?

xu-cheng:
Fixed by #23

PR #16, opened 2017-04-05
Fix build on 10.4 and i386

mistydemeo:
This adapts a few fixes already in use on Tigerbrew.

### DIFF
--- a/Formula/portable-openssl.rb
+++ b/Formula/portable-openssl.rb
@@ -16,6 +16,12 @@ class PortableOpenssl < PortableFormula
     sha256 "0f119da204025da7808273fab42ed8e030cafb5c7ea4e1deda4e75f066f528fb"
   end
 
+  # Fixes ASM for i386 builds on older OS Xs
+  patch :p0 do
+    url "https://trac.macports.org/export/144472/trunk/dports/devel/openssl/files/x86_64-asm-on-i386.patch"
+    sha256 "98ffb308aa04c14db9c21769f1c5ff09d63eb85ce9afdf002598823c45edef6d"
+  end
+
   def openssldir
     libexec/"etc/openssl"
   end
@@ -37,7 +43,7 @@ class PortableOpenssl < PortableFormula
   end
 
   def configure_args
-    %W[
+    args = %W[
       --prefix=#{prefix}
       --openssldir=#{openssldir}
       no-ssl2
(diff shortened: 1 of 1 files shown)


Formula/portable-openssl.rb
```diff
@@ -37,14 +42,18 @@ def arch_args
...
+    args = %W[
       --prefix=#{prefix}
       --openssldir=#{openssldir}
       no-ssl2
       zlib-dynamic
       no-shared
       enable-cms
     ]
+
+    args << "no-asm" if MacOS.version < :leopard
```
mistydemeo: Xcode 2.5's assembler is too old, and I'm not yet at the point of backporting a new assembler.

Formula/portable-openssl.rb
```diff
@@ -16,6 +16,11 @@ class PortableOpenssl < PortableFormula
     sha256 "0f119da204025da7808273fab42ed8e030cafb5c7ea4e1deda4e75f066f528fb"
   end
 
+  patch :p0 do
```
mistydemeo: Necessary for i386 builds.

Formula/portable-openssl.rb
```diff
@@ -74,7 +83,7 @@ def install
       system "make"
       system "make", "test"
 
-      if build.with? "universal"
+      if build.with? "universal" && archs.length > 1
```
MikeMcQuaid: Probably needs parens here to work as expected.

Formula/portable-openssl.rb
```diff
@@ -83,7 +92,7 @@ def install
     system "make", "install", "MANDIR=#{man}"
     rm_rf man
 
-    if build.with? "universal"
+    if build.with? "universal" && archs.length > 1
```
MikeMcQuaid: Probably needs parens here to work as expected.

Formula/portable-openssl.rb
mistydemeo: Oh, good point.

I'm also rethinking my universal approach - even if "universal" usually doesn't include 64-bit archs on Tiger and Leopard, I may look at making the portable formulae build i386/ppc.

xu-cheng: Just my two cents, this could be just `if archs.length > 1`

mistydemeo:
Removed the universal commit.

PR #17, opened 2017-04-05
portable-formula: enable i386/ppc universal on Tigerbrew

mistydemeo:
This replaces the awkward universal arch handling in #16.

Tigerbrew ordinarily builds 32-bit only on 10.4 and 10.5, even on 64-bit hardware. As a result, "universal" builds are still single-arch. The reason for this is that the tooling is unfinished on Tiger, and the linker is fiddly on 10.5, making 64-bit builds an error-prone affair. However, for these portable formulae, I'd love to support i386/ppc universal - that way I can build one binary, on one machine, that works on both Intel and PowerPC. The Apple-provided gcc-4.0 and gcc-4.2 are able to do this on both architectures.

cc @xu-cheng

### DIFF
--- a/Abstract/portable-formula.rb
+++ b/Abstract/portable-formula.rb
@@ -9,6 +9,15 @@ module PortableFormulaMixin
       EOS
     end
 
+    # Overrideable per-formula, but try to make sure our universal
+    # arches make it into the environment.
+    # This is important because in some environments (e.g. 10.4/10.5)
+    # our arches differ from the usual defaults.
+    if OS.mac? && build.with? "universal"
+      ENV.permit_arch_flags
+      ENV.append_to_cflags archs.map {|a| "-arch #{a}"}.join(" ")
+    end
+
     if OS.linux?
       # reset Linuxbrew env, because we want to build formula against
       # libraries offered by system (CentOS docker) rather than Linuxbrew.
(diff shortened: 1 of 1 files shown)


Abstract/portable-formula.rb
```diff
@@ -49,6 +58,21 @@ def archs
...
     end
 
+    # On Tiger and Leopard, override the default behaviour.
+    # Normally we don't build 64-bit there, because the linker is
+    # temperamental, and so "universal" builds don't usually happen.
+    # However, for the purposes of distributing portable packages,
+    # it's very useful to be able to build i386/ppc binaries for use
+    # on both Intel and PowerPC Macs. The Apple-provided compilers
+    # are capable of this on both Intel and Mac hosts.
+    if MacOS.version < :snow_leopard
```
xu-cheng: You probably want to add `OS.mac?` as formulae in this repo should work on both mac and linux.

Abstract/portable-formula.rb
```diff
@@ -49,6 +58,21 @@ def archs
...
+      if build.with? "universal"
+        return [:i386, :ppc]
+      else
+        return [Hardware::CPU.arch_32_bit]
```
xu-cheng: It seems that two different styles are used here:
* `Hardware::CPU.universal_archs/Hardware::CPU.arch_64_bit/Hardware::CPU.arch_32_bit` 
* `:x86_64/:i386/:ppc`

I wonder would it better to enforce only one style here. At least, this will make it easy to read.

Abstract/portable-formula.rb
mistydemeo: Good idea, done.

Abstract/portable-formula.rb
mistydemeo: I hardcoded here because we don't have helpers for "give me both 32-bit arches for PPC-era Macs"; it's a niche enough usecase I don't think we need it as a general case.

PR #20, opened 2017-04-06
portable-git various improvement

xu-cheng:
* Guard macos version check with OS.mac?
* use SHA256


xu-cheng:
> Do we care about making MacOS checks into OS::Mac checks?

I am fine with both. And it seems that `MacOS` is used more frequently.

PR #23, opened 2017-04-07
Fix broken zlib on Linux

xu-cheng:
Replace #11


sjackman:
Hi, Xu. Is this PR addressing an issue that you experienced? Or is the primary intention to achieve static linking against zlib?

xu-cheng:
@sjackman currently portable openssl build is broken on the Linux. This PR is intended to solve this.

PR #25, opened 2017-04-08
portable-curl: fix for Linux

xu-cheng:
Fix portable-c-ares for Linux by linking to c-ares.


sjackman:
Thanks for this work, Xu.

PR #26, opened 2017-04-08
portable-formula: fix (potential?) typo in comment

alyssais:
I'm guessing this is what you meant, @mistydemeo? I know very little about this though so could be wrong.


xu-cheng:
Thanks @alyssais

mistydemeo:
Good catch, thanks!

PR #27, opened 2017-04-09
Add portable-ruby@2.2

mistydemeo:
This adds a formula for Ruby 2.2, the newest version that currently builds universal when statically linked. Ruby 2.3 and 2.4 currently don't build [due to a bug with statically linked ext](https://bugs.ruby-lang.org/issues/13413). This also includes a patch that scopes the universal workaround to only be applied to `portable-ruby`, not other Ruby versions.

Universal is important for me with Tigerbrew, as I'll want to be shipping a single binary for both PowerPC and Intel. I also prefer shipping a version that's currently supported upstream; 2.0 and 2.1 are both EOL.

### DIFF
--- a/Abstract/portable-formula.rb
+++ b/Abstract/portable-formula.rb
@@ -43,8 +43,8 @@ class PortableFormula < Formula
     subclass.class_eval do
       keg_only "Portable formula is keg-only."
 
-      # TODO remove `subclass.name !~ /PortableRuby/` when updating portable-ruby to 2.1 or above.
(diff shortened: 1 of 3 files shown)


xu-cheng:
Could you apply similar change as https://github.com/Homebrew/homebrew-portable/commit/7c7439c01c4b75f29efa12f780bd14bcabe8d01d for Linux build and update README table to reflect this new formula? Thanks.  :+1:

mistydemeo:
Done!

PR #28, opened 2017-04-10
portable-ruby@2.2: fix build on PPC and Tiger

mistydemeo:
* Fix access to the `mcontext` struct, used on Intel, which changed shape between 10.4 and 10.5.
* Fix the universal `CFLAGS` by ensuring GCC doesn't interpret `-march` as an argument to be used when `-arch` is `ppc`.
* Tiger's make (3.80) is too old to understand Ruby's makefile; 3.81 or later is needed.

### DIFF
--- a/Formula/portable-ruby@2.2.rb
+++ b/Formula/portable-ruby@2.2.rb
@@ -6,6 +6,7 @@ class PortableRubyAT22 < PortableFormula
   url "https://cache.ruby-lang.org/pub/ruby/2.2/ruby-2.2.7.tar.gz"
   sha256 "374184c6c5bbc88fb7bad422368d4053a236fb6587f0eff76146dcba57f93da5"
 
+  depends_on "make" => :build if OS.mac? && MacOS.version < :leopard
   depends_on "makedepend" => :build
   depends_on "pkg-config" => :build
   depends_on "portable-readline" => :build
@@ -90,8 +91,8 @@ class PortableRubyAT22 < PortableFormula
     args << "--with-opt-dir=#{paths.join(":")}"
 
     system "./configure", *args
-    system "make"
-    system "make", "install"
+    make
+    make "install"
 
     abi_version = `#{bin}/ruby -rrbconfig -e 'print RbConfig::CONFIG["ruby_version"]'`
     abi_arch = `#{bin}/ruby -rrbconfig -e 'print RbConfig::CONFIG["arch"]'`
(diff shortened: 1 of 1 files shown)


Formula/portable-ruby@2.2.rb
```diff
@@ -6,6 +6,7 @@ class PortableRubyAT22 < PortableFormula
   url "https://cache.ruby-lang.org/pub/ruby/2.2/ruby-2.2.7.tar.gz"
   sha256 "374184c6c5bbc88fb7bad422368d4053a236fb6587f0eff76146dcba57f93da5"
 
+  depends_on "make" => :build if MacOS.version < :leopard
```
xu-cheng: Need `OS.mac?` here.

Formula/portable-ruby@2.2.rb
```diff
@@ -17,6 +18,27 @@ class PortableRubyAT22 < PortableFormula
   end
 
   def install
+    # mcontext types had a member named `ss` instead of `__ss`
+    # prior to Leopard; see
+    # https://github.com/mistydemeo/tigerbrew/issues/473
+    if Hardware::CPU.intel? && MacOS.version < :leopard
```
xu-cheng: Need `OS.mac?` here.

Formula/portable-ruby@2.2.rb
```diff
@@ -64,8 +89,8 @@ def install
     args << "--with-opt-dir=#{paths.join(":")}"
 
     system "./configure", *args
-    system "make"
-    system "make", "install"
+    make
+    make "install"
```
xu-cheng: Should we use `make` method for all of formulae in this repo?

Formula/portable-ruby@2.2.rb
xu-cheng: If this is also needed for other formulae,  we could move `depends_on "make" => :build` to `PortableFormulae`

mistydemeo: Right now, it's only needed for Rubies, not other formulae.

Formula/portable-ruby@2.2.rb
mistydemeo: In my experience, 99% of things work on make 3.80; only a few things run into its bugs.

xu-cheng: Cool. I just thought using `make` method everywhere may be more consistence.

mistydemeo: If we run into another non-ruby formula in this tap that needs it, I'll convert everything over!

PR #29, opened 2017-04-13
tweak LDFLAGS

xu-cheng:
* Enable `-Wl,search_paths_first` for all formulae on macOS to prefer to
  using our portable static libraries.
* Do not pass `-Wl,search_paths_first` on Linux. It is an invalid flag for
  gcc, which always prefers to paths specified by `-L` already.
* Refactor `PortableFormulaMixin#install`.


xu-cheng:
@mistydemeo As I don't have access to Mac OS X 10.5 and 10.4. Would you mind to test this PR, so I didn't accidentally break anything? Thanks 🙇

PR #33, opened 2017-04-15
curl: fix replacing CFLAGS

mistydemeo:
curl sets `-mmacosx-version-min=10.8` in CFLAGS if not otherwise specified, and since we're overriding CFLAGS, the value set by stdenv gets erased. This does a more conservative replacement, which only strips arch-related flags.


mistydemeo:
I will eventually get superenv going on Tiger, but

a) it will require more work because Tiger doesn't ship `xcrun`, and
b) I need to upgrade Ruby for all Tigerbrew users, e.g. the formulae in this repo are a prerequisite.

PR #34, opened 2017-04-15
portable-git: needs openssl on Tiger/Leopard

mistydemeo:
Tigerbrew already does this; the version on Tiger is too old to build git, and Leopard's is realistically too old to want to use as well. While CommonCrypto is shipped on Leopard, I trust modern OpenSSL more than I trust 10-year-old CommonCrypto.

### DIFF
--- a/Formula/portable-git.rb
+++ b/Formula/portable-git.rb
@@ -16,6 +16,7 @@ class PortableGit < PortableFormula
     depends_on "portable-expat" => :build
   end
 
+  depends_on "portable-openssl" => :build
   depends_on "portable-zlib" => :build if OS.linux?
 
   # ld64 understands -rpath but rejects it on Tiger
@@ -57,7 +58,6 @@ class PortableGit < PortableFormula
     ENV["NO_PERL_MAKEMAKER"] = "1"
     ENV["NO_GETTEXT"] = "1"
     ENV["NO_TCLTK"] = "1"
-    ENV["NO_OPENSSL"] = "1"
     if OS.linux? || OS::Mac.version < :leopard
       ENV["EXPATDIR"] = expat.opt_prefix
     end


xu-cheng:
Is this really necessary?  As far as I'm understanding, currently git is using ssl implemention from portable-curl, which already packs portable-openssl already.

mistydemeo:
Yes, it tries to link directly to openssl and fails due to not being able to find the correct symbols.

mistydemeo:
Specifically, it looks like it's getting `-lssl -lcrypto` from `curl-config --static-libs`, so it still tries to link directly against openssl: https://gist.github.com/025a47471b8057c2f0734605a575e6e9

It returns the flags `/usr/local/Cellar/curl/7.50.3/lib/libcurl.a -L/usr/local/lib -Wl,-headerpad_max_install_names -L/usr/local/Cellar/openssl/1.0.2j/lib -lssl -lcrypto -lssl -lcrypto -lldap -lz`.

xu-cheng:
> it looks like it's getting -lssl -lcrypto from curl-config --static-libs

Yes, it is the case. 

Since we always linking to openssl for all the OS (modern or old mac and linux), I think we should just include portable-openssl unconditionally, like this:

```diff
diff --git a/Formula/portable-git.rb b/Formula/portable-git.rb
index 522800f..894c970 100644
--- a/Formula/portable-git.rb
+++ b/Formula/portable-git.rb
@@ -11,6 +11,7 @@ class PortableGit < PortableFormula
     depends_on "gnu-tar" => :build
   end
 
+  depends_on "portable-openssl" => :build
   depends_on "portable-curl" => :build
   if OS.linux? || OS::Mac.version < :leopard
     depends_on "portable-expat" => :build
@@ -56,7 +57,6 @@ class PortableGit < PortableFormula
     ENV["NO_PERL_MAKEMAKER"] = "1"
     ENV["NO_GETTEXT"] = "1"
     ENV["NO_TCLTK"] = "1"
-    ENV["NO_OPENSSL"] = "1"
     if OS.linux? || OS::Mac.version < :leopard
       ENV["EXPATDIR"] = expat.opt_prefix
     end
```

mistydemeo:
Updated to link against openssl unconditionally.

PR #35, opened 2017-04-16
ARM support

xu-cheng:
Requiring: 
 * [x] https://github.com/Linuxbrew/brew/pull/350
 * [x] https://github.com/Linuxbrew/brew/pull/351

### DIFF
--- a/Formula/portable-openssl.rb
+++ b/Formula/portable-openssl.rb
@@ -39,6 +39,7 @@ class PortableOpenssl < PortableFormula
       {
         :x86_64 => %w[linux-x86_64],
         :i386  => %w[linux-generic32],
+        :arm => %w[linux-armv4],
       }
     end
   end
--- a/README.md
+++ b/README.md
@@ -8,33 +8,41 @@ Just `brew install homebrew/portable/<formula>`.
 ### macOS
 Run `brew portable-package <formula>`. Ideally inside an OS X 10.5 VM so it is compatible with old macOS versions.
 
-### Linux
+### Linux (x86_64)
 Run `brew portable-package <formula>`. Ideally this should be run inside the CentOS 5 Docker container with:
 ```bash
-docker build -f docker/Dockerfile -t homebrew-portable:latest .
-docker run -t -i homebrew-portable /bin/bash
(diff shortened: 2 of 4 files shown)


MikeMcQuaid:
Can this be force-pushed to trigger CI? Thanks!

xu-cheng:
This PR is targeted on Linuxbrew only and currently work in progress. So it would not affect CI status, which is for macOS. 

I will trigger CI when it is ready.

Formula/portable-openssl.rb
```diff
@@ -39,6 +39,7 @@ def arch_args
       {
         :x86_64 => %w[linux-x86_64],
         :i386  => %w[linux-generic32],
+        :arm => %w[linux-armv4],
```
sjackman: Should this be `armv6` for Raspberry Pi version 1? What's the minimum ARM CPU that you plan to support?

xu-cheng: This is the only flag OpenSSL offers for ARM. 

As the minimum CPU, maybe ARMv7 as that is what the oldest docker image of Raspberry PI I obtained uses.

sjackman: It must be no newer than ARMv6, as the Raspberry Pi 1 has an ARMv6 processor.
See https://en.wikipedia.org/wiki/Raspberry_Pi#Processor

xu-cheng: Ok, that should be much a problem considering we pass `-march=armv6` flag for compiling.
However since ARMv7 is the oldest docker image, I wouldn't be able to test it for ARMv6. 

@sjackman If you have access to any Raspberry Pi hardware, it would be great if you can help for testing the building results on the real device.

xu-cheng:
Merged as it is for now. This is the first step to bring portable tools for ARM.

Formula/portable-openssl.rb
sjackman: I do have a Raspberry Pi 1, but limited free time at home with two kids about. Once you think it's ready for a test on real hardware, I can take a look at it.

xu-cheng: Great. I will report back to you once I have done extensive test on my VM.

PR #37, opened 2017-04-17
Add Travis CI

MikeMcQuaid:
This is necessary to ensure changes to formulae in this repo work as expected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+language: ruby
+os: osx
+env: OSX=10.12
+osx_image: xcode8.1
+rvm: system
+before_install:
+  - export TRAVIS_COMMIT="$(git rev-parse --verify -q HEAD)"
+  - if [ -f ".git/shallow" ]; then
+      travis_retry git fetch --unshallow;
+    fi
+  - HOMEBREW_REPOSITORY="$(brew --repo)"
+  - sudo chown -R "$USER" "$HOMEBREW_REPOSITORY"
+  - git -C "$HOMEBREW_REPOSITORY" reset --hard origin/master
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,26 @@
+language: ruby
+os: osx
+env: OSX=10.12
```
mistydemeo: Given that we recommend building formulae in this repo on older versions of OS X, maybe we should test using an older release too.

MikeMcQuaid: @mistydemeo Travis CI has limited numbers of workers running at once per org and a generally limited number of macOS workers. As a result I think it's better sticking with a single version but I'm 👍 to this being an older one. Just wanted to 🚢 this to test the other PRs in progress. Feel free to change in a new PR 👍

xu-cheng: We should just run CI on the oldest available macos image.

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+  - git -C "$HOMEBREW_REPOSITORY" reset --hard origin/master
+  - brew update || brew update
+  - HOMEBREW_TAP_DIR="$(brew --repo "$TRAVIS_REPO_SLUG")"
+  - mkdir -p "$HOMEBREW_TAP_DIR"
+  - rm -rf "$HOMEBREW_TAP_DIR"
+  - ln -s "$PWD" "$HOMEBREW_TAP_DIR"
+  - export HOMEBREW_DEVELOPER="1"
+  - ulimit -n 1024
+script:
+  - brew test-bot
```
xu-cheng: This should be `brew tests --skip-relocation`

PR #39, opened 2017-04-20
rewrite portable-package cmd

xu-cheng:
* Better help message.
* Remove obsoleted ruby version check. Homebrew should always run with
  Ruby 2.0+ now. At the same time, old ruby check won't work as it will
  just crash when loading formula if Homebrew is somehow run with Ruby
  1.8.
* Remove obsoleted HOMEBREW_PREFER_64_BIT. We start to build 32bit formulae
  on OS X older than Snow Leopard since 6a3a192a65dfa16636525aa90438e9164ba5e9c0.
* Add new option to not uninstall deps, useful for developing.
* Use `Utils.popen_read` instead of pipe to enforce proper shell esecape.
  Otherwise, `brew portable-package some_shell_code_injection` will :boom:.
* Use `brew linkage` on both Mac and Linux. Previously,
  `ldd #{f.bin}/#{name.gsub(/^portable-/, "")}` will break for non elf
  file.
* Use `safe_system`, instead of superfluous `puts Utils.popen_read`.
* Support various bottle options.
* Reuse existing bottles when building multiple formulae.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_install:
   - ulimit -n 1024
 script:
   - brew test-bot --skip-relocation
-  - brew portable-package portable-ruby
+  - brew portable-package --write portable-ruby
 notifications:
   email:
     on_success: never
--- a/cmd/brew-portable-package.rb
+++ b/cmd/brew-portable-package.rb
@@ -1,28 +1,54 @@
-# Build a portable Homebrew utility bottle.
-#
-# Usage: brew portable-package <formula> [formula...]
+#: `brew` `portable-package` [`--no-uninstall-deps`] [`--no-rebuild`|`--keep-old`] [`--write` [`--no-commit`]] <formulae>:
+#:    Build and package portable formulae.
+#:
+#:    Unless `--uninstall-deps` is passed, all dependencies of portable
+#:    formulae will be uninstalled before test. Useful for developing purpose.
+#:
+#:    If the formula specifies a rebuild version, it will be incremented in the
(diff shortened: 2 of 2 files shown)


MikeMcQuaid:
Could we get some `travis.yml` tests that exercise this command (in a separate PR that can be merged before this one)? Thanks!

MikeMcQuaid:
To explain further the motivation: this is a repository that Homebrew relies on for core infrastructure (building Ruby). If we need to update that in future at short notice due to critical security vulnerabilities I'm concerned that the amount of changes in this repository recently and in this PR may cause bugs that render us unable to do that.

xu-cheng:
> Could we get some travis.yml tests

Done in this PR. (So it will also test new flag and apply bug fix)

> I'm concerned that the amount of changes in this repository recently and in this PR may cause bugs that render us unable to do that.

I understand your concern. Just want to mention that all recent changes are either:
* Important bug fixes.
* Improvement for previous missing support. e.g. PowerPC for Tigerbrew and ARM for Linuxbrew.

MikeMcQuaid:
> Done in this PR. (So it will also test new flag and apply bug fix)

Can you do so in another PR? I'd like to verify the tests work both before and after this change, thanks.



> I understand your concern. Just want to mention that all recent changes are either:
> 
> Important bug fixes.
> Improvement for previous missing support. e.g. PowerPC for Tigerbrew and ARM for Linuxbrew.

Sure, they are fine, just none are needed for Homebrew itself.

PR #41, opened 2017-07-07
Add formulae for portable-ruby 2.3 and 2.4

mistydemeo:
These were broken due to an [upstream bug](https://bugs.ruby-lang.org/issues/13413) in makefile generation, which prevented the `.rb` portions of statically linked extensions from being installed. This has been fixed in trunk, so I've backported the relevant patch in these formulae and put in a request for it to be backported into the branches upstream.

Trunk, and by extension the upcoming 2.5, is still broken in statically-linked builds for unrelated reasons. I'm looking into the issue.

Having these work will be useful for us since we want to start shipping a vendored Ruby 2.3 when 10.13 is out. :)

### DIFF
--- a/Formula/portable-ruby@2.3.rb
+++ b/Formula/portable-ruby@2.3.rb
@@ -1,0 +1,139 @@
+require File.expand_path("../../Abstract/portable-formula", __FILE__)
+
+class PortableRubyAT23 < PortableFormula
+  desc "Portable ruby 2.3"
+  homepage "https://www.ruby-lang.org/"
+  # This isn't the latest 2.3.3, but it matches the version shipped
+  # in current betas of macOS 10.13.
+  url "https://cache.ruby-lang.org/pub/ruby/2.3/ruby-2.3.3.tar.bz2"
+  sha256 "882e6146ed26c6e78c02342835f5d46b86de95f0dc4e16543294bc656594cc5b"
+
+  depends_on "make" => :build if OS.mac? && MacOS.version < :leopard
+  depends_on "makedepend" => :build
+  depends_on "pkg-config" => :build
+  depends_on "portable-readline" => :build
+  depends_on "portable-libyaml" => :build
+  depends_on "portable-openssl" => :build
+  if OS.linux?
+    depends_on "portable-ncurses" => :build
+    depends_on "portable-zlib" => :build
+  end
+
+  # Fixes the static build: https://bugs.ruby-lang.org/issues/13413
+  # This has been backported into the 2.3 branch, but isn't in the
+  # release we're building.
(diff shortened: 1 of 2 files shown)


reitermarkus:
<a href="#" title="+1"></a>

xu-cheng:
Should we remove old rubies and only keep ruby 2,3?

Otherwise, :+1:

Formula/portable-ruby@2.3.rb
```diff
@@ -0,0 +1,135 @@
...
+  depends_on "portable-libyaml" => :build
+  depends_on "portable-openssl" => :build
+  if OS.linux?
+    depends_on "portable-ncurses" => :build
+    depends_on "portable-zlib" => :build
+  end
+
+  # Fixes the static build: https://bugs.ruby-lang.org/issues/13413
+  patch do
+    url "https://github.com/ruby/ruby/commit/b3dbeb6e90f316584f70e33f6bfb9d83fa5f30d3.patch?full_index=1"
```
xu-cheng: Do we need `?full_index=1`

MikeMcQuaid:
Great work figuring out the bug here 👏 

> portable-ruby@2.3 2.3.4

I think we should match Apple's version here: `ruby 2.3.3p222 (2016-11-21 revision 56859)`.

> portable-ruby@2.4 2.4.1

Where are we planning on using this formula?

Formula/portable-ruby@2.3.rb
mistydemeo: Yeah, this is a transitional feature for raw patch URLs that uses unabbreviated git refs instead of shortened ones.

mistydemeo:
> I think we should match Apple's version here: ruby 2.3.3p222 (2016-11-21 revision 56859).

👍 

> Where are we planning on using this formula?

It's mainly here for us to experiment with. Given the issues we've seen with static builds in the past, I also think it's good for us to test newer Ruby versions by having formulae for them here, even if we're not using them yet.

MikeMcQuaid:
> It's mainly here for us to experiment with. Given the issues we've seen with static builds in the past, I also think it's good for us to test newer Ruby versions by having formulae for them here, even if we're not using them yet.

Cool with me, just curious 👍. I think at some point it would be worth separating this repo into the stuff we're using/supporting in production and the stuff we aren't (but that's not for this PR).

mistydemeo:
Updated to switch portable-ruby@2.3 to use Ruby 2.3.3 - that point release is the same patchlevel as macOS 10.13's version. Also updated the comment on 2.3's patch to indicate that the commit has been backported onto the 2.3 branch.

PR #42, opened 2017-07-18
Abstract/portable-formula: Tweak keg_only reason

rwhogg:
Make it lowercase and not end with a period, because
it fits in with the caveats sentece better that way.

Signed-off-by: Bob W. Hogg <rwhogg@linux.com>


MikeMcQuaid:
Thanks @rwhogg!

PR #44, opened 2017-09-15
Fix portable-ruby@2.3 on Leopard

mistydemeo:
This uses an upstream patch to fix the Leopard build.

It also restricts a flags change to stdenv; it's not necessary on superenv, and uses a method we don't expose in superenv either.

I've also promoted 2.3 to being the default portable Ruby now that 10.13 is about to come out with that as its system Ruby.


Formula/portable-ruby.rb
```diff
@@ -1,17 +1,14 @@
 require File.expand_path("../../Abstract/portable-formula", __FILE__)
 
-class PortableRuby < PortableFormula
-  desc "Portable ruby"
+class PortableRubyAT23 < PortableFormula
```
reitermarkus: This should stay at `PortableRuby`, right?

mistydemeo: Oops, you're right - fixed.

Formula/portable-ruby.rb
```diff
@@ -3,15 +3,12 @@
...
-  sha256 "087ad4dec748cfe665c856dbfbabdee5520268e94bb81a1d8565d76c3cc62166"
-
-  bottle do
-    cellar :any_skip_relocation
-    sha256 "5c1240abe4be91c9774a0089c2a38a8ccfff87c009e8e5786730c659d5e633f7" => :leopard_64
-    sha256 "dbb5118a22a6a75cc77e62544a3d8786d383fab1bdaf8c154951268807357bf0" => :x86_64_linux
-  end
+  # This isn't the latest 2.3.3, but it matches the version shipped
+  # in current betas of macOS 10.13.
+  url "https://cache.ruby-lang.org/pub/ruby/2.3/ruby-2.3.3.tar.bz2"
```
MikeMcQuaid: I wonder if we want to have the `p222` as part of the `version`?

Formula/portable-ruby.rb
```diff
@@ -3,15 +3,12 @@
...
-  url "https://cache.ruby-lang.org/pub/ruby/2.0/ruby-2.0.0-p648.tar.bz2"
-  sha256 "087ad4dec748cfe665c856dbfbabdee5520268e94bb81a1d8565d76c3cc62166"
-
-  bottle do
-    cellar :any_skip_relocation
-    sha256 "5c1240abe4be91c9774a0089c2a38a8ccfff87c009e8e5786730c659d5e633f7" => :leopard_64
-    sha256 "dbb5118a22a6a75cc77e62544a3d8786d383fab1bdaf8c154951268807357bf0" => :x86_64_linux
-  end
+  # This isn't the latest 2.3.3, but it matches the version shipped
+  # in current betas of macOS 10.13.
```
MikeMcQuaid: or latest GM.

MikeMcQuaid:
@mistydemeo Got this building (and working) on the Leopard VM but not sure if it's working in the CI environment yet.

MikeMcQuaid:
@mistydemeo @xu-cheng I seem to have hit a problem here I now have no idea how to work around:
- 32-bit-only Ruby won't build native extensions on 64-bit macOS
- Universal Ruby will build universal, native extensions on 64-bit macOS but can't load them
- Cannot build 64-bit-only Ruby on 10.5 Leopard at all

Thoughts?

Formula/portable-ruby.rb
```diff
@@ -33,18 +67,18 @@ def install
       --prefix=#{prefix}
       --enable-load-relative
       --with-static-linked-ext
-      --disable-dln
```
mistydemeo: Nice! It's still statically linked with this?

MikeMcQuaid: Yes and with it specified you cannot load any native extensions at runtime 😨

Abstract/portable-formula.rb
```diff
@@ -48,27 +48,21 @@ def self.inherited(subclass)
     subclass.class_eval do
       keg_only "portable formulae are keg-only"
 
-      # TODO remove `subclass.name !~ /PortableRuby$/` when updating portable-ruby to 2.1 or above.
-      option "without-universal", "Don't build a universal binary" if OS.mac? && subclass.name !~ /PortableRuby$/
+      option "without-universal", "Don't build a universal binary"
```
sjackman: To prevent running `lipo` on Linux, which are guarded by `if build.with? "universal"`, change this line to:
```ruby
option "without-universal", "Don't build a universal binary" if OS.mac?
```

PR #45, opened 2017-09-19
portable-openssl 1.0.2l

MikeMcQuaid:
Extracted from https://github.com/Homebrew/homebrew-portable/pull/44.


MikeMcQuaid:
Moving back into #44.

PR #46, opened 2017-09-19
portable-ruby: Add http mirror

sjackman:
Fix the error:
```
```


MikeMcQuaid:
Thanks @sjackman!

PR #48, opened 2017-09-21
portable-openssl: Add http mirror

sjackman:
Fix the error:
```
==> Downloading from https://curl.haxx.se/ca/cacert-2017-01-18.pem
```

Note that http://curl.haxx.se redirects to https://curl.haxx.se


MikeMcQuaid:
Thanks @sjackman!

PR #50, opened 2017-09-23
portable-ruby: enable coverage.

MikeMcQuaid:
Do this for e.g. `simplecov`.


MikeMcQuaid:
@Homebrew/linux @sjackman Without this you cannot `brew tests --coverage`. You can feel free to make a new build or not depending on your requirements but I'd advise decide either way before merging the next tag to Linuxbrew.

sjackman:
I'll build a new bottle. Thanks for the heads up.

sjackman:
Should this be a `revision 1` rather than a `rebuild 1`?

MikeMcQuaid:
@sjackman I think a `rebuild 1` is fine in this relatively rare case because we're not wanting to force upgrades.

sjackman:
I've built the `portable-ruby` 2.3.3 rebuild 1 bottle for Linux.
https://github.com/Homebrew/brew/pull/3326

PR #51, opened 2017-11-13
portable-ruby: update to track the 2.3 maintenance branch

mistydemeo:
This is based on a conversation with @JCount. There are security and other bugfixes included in the 2,3 branch past the 2.3.3 release shipped by Apple; it feels better to ship a Ruby without known issues, even if that means deviating from Apple's Ruby a bit.

This PR switches over to building from the 2.3 maintenance branch. As a result, I was able to remove all the patches currently in the formula; they're all already in that branch.

### DIFF
--- a/Formula/portable-ruby.rb
+++ b/Formula/portable-ruby.rb
@@ -17,6 +17,7 @@ class PortableRuby < PortableFormula
   end
 
   depends_on "autoconf" => :build
+  depends_on "bison" => :build if OS.mac? && MacOS.version < :leopard
   depends_on "make" => :build if OS.mac? && MacOS.version < :leopard
   depends_on "makedepend" => :build
   depends_on "pkg-config" => :build
--- a/Formula/portable-ruby.rb
+++ b/Formula/portable-ruby.rb
@@ -3,10 +3,12 @@ require File.expand_path("../../Abstract/portable-formula", __FILE__)
 class PortableRuby < PortableFormula
   desc "Portable ruby"
   homepage "https://www.ruby-lang.org/"
-  # This isn't the latest 2.3.3 but is the version shipped in macOS 10.13.
-  url "https://cache.ruby-lang.org/pub/ruby/2.3/ruby-2.3.3.tar.bz2"
-  mirror "http://cache.ruby-lang.org/pub/ruby/2.3/ruby-2.3.3.tar.bz2"
-  sha256 "882e6146ed26c6e78c02342835f5d46b86de95f0dc4e16543294bc656594cc5b"
+  # Tracks the 2.3 maintenance branch.
+  # We don't need to update for every new commit in this branch;
+  # we should update when something important comes up though.
+  url "https://github.com/ruby/ruby.git",
+    :revision => "463a56d14b9241c8631d71d286760ea092d70c36"
+  version "2.3-2017-09-15"
 
   bottle do
     cellar :any_skip_relocation
@@ -14,6 +16,7 @@ class PortableRuby < PortableFormula
     sha256 "34ce9e4c9c1be28db564d744165aa29291426f8a3d2ef806ba4f0b9175aedb2b" => :leopard_64_or_later
   end
 
+  depends_on "autoconf" => :build
   depends_on "make" => :build if OS.mac? && MacOS.version < :leopard
   depends_on "makedepend" => :build
   depends_on "pkg-config" => :build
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> There are security and other bugfixes included in the 2,3 branch past the 2.3.3 release shipped by Apple; it feels better to ship a Ruby without known issues, even if that means deviating from Apple's Ruby a bit.

I very strongly disagree, I'm afraid. The intention of our portable ruby is to avoid supporting multiple Ruby versions so we should distribute as closely identical version to the one Apple does. Additionally, the portable-ruby upgrade process is not particularly nice or smooth and I'd rather avoid any upgrades where not warranted by issues directly affecting Homebrew itself.

Formula/portable-ruby.rb
```diff
@@ -117,14 +96,12 @@ def install
...
 
+    system "autoconf"
+
     system "./configure", *args
     make
     make "install"
 
-    # rake is a binstub for the RubyGem in 2.3 and has a hardcoded PATH.
-    # We don't need the binstub so remove it.
-    rm bin/"rake"
```
MikeMcQuaid: Is this fixed upstream? If so, can we backport the patch instead?

MikeMcQuaid:
https://github.com/mistydemeo/homebrew-portable/blob/b4f38e6fc30c94aa122bb98b1ea28e51f60c1865/Formula/portable-ruby.rb#L112-L120 

It's probably worth changing this at some point to a solution that doesn't e.g. set `SSL_CERT_FILE` if you `require "securerandom"` as this breaks `curl` when called by Ruby subprocesses.

MikeMcQuaid:
> The intention of our portable ruby is to avoid supporting multiple Ruby versions so we should distribute as closely identical version to the one Apple does

With that in mind: https://opensource.apple.com/source/ruby/ruby-131/ruby/ may be the best upstream for us to use.

Formula/portable-ruby.rb
mistydemeo: Looks like the binary isn't shipped anymore (?) - at least it wasn't there when I tried to delete it. I'll see if I can reproduce, and see if there's a particular commit I can grab.

MikeMcQuaid: @mistydemeo Probably worth talking through the other comments first to make sure it's worth your time.

JCount:
Currently, there are widely known security vulnerabilities in the version of `portable-ruby` we ship. These issues, among others, have been addressed in trunk and backported to the `ruby_2_3` branch. These backports are cherry-picked to maintain both API and ABI compatibility in accordance with [ruby's version policy](https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/) post version `2.1.0`, _theoretically_ meaning that `Ruby 2.3.5` should be an almost drop-in replacement.

[Introducing a semantic versioning scheme and branching policy](https://bugs.ruby-lang.org/issues/8835)

Given that we have already switched to essentially requiring users to use our `portable-ruby`, I cannot see how we can merely match Apple's version, using an outdated, vulnerable `TEENY` version given:
>(3) TEENY version is incremented when functionality is added in a backward-compatible manner.

Being cognizant of these security issues and how at least some of them apply to our use case, I think it would be good to address them, given the ability to do so.

-----

>The intention of our portable ruby is to avoid supporting multiple Ruby versions so we should distribute as closely identical version to the one Apple does.

@MikeMcQuaid What do you consider to be closely identical "enough", as in what sort of criteria do you use, and how much and what sort of patching do you think is appropriate?

MikeMcQuaid:
> Currently, there are widely known security vulnerabilities in the version of portable-ruby we ship.

Which vulnerabilities? Please demonstrate an exploit of one of these vulnerabilities through use of Homebrew. 

> theoretically meaning that Ruby 2.3.5 should be an almost drop-in replacement.

Theoretically is not good enough. Additionally, where there are bugs in the system Ruby that require workarounds: we want to workaround them everywhere.

> Given that we have already switched to essentially requiring users to use our portable-ruby

We have not. This is only the case on people not using High Sierra. The number of people using High Sierra will increase over time. In under two years Apple may not have updated the system Ruby again so we may end up with no supported versions using portable-ruby.

> Being cognizant of these security issues and how at least some of them apply to our use case, I think it would be good to address them, given the ability to do so.

Again: which security issues and please demonstrate how they are exploitable in Homebrew's case?

Additionally, if these vulnerabilities are present in our Ruby but not in Apple's Ruby: we should apply the same patches Apple has, I agree. If they are present in Apple's Ruby: it's not sufficient to ship a portable Ruby with these issues fixed as some users will still be using the system Ruby.

> @MikeMcQuaid What do you consider to be closely identical "enough", as in what sort of criteria do you use, and how much and what sort of patching do you think is appropriate?

I've explained my preferences above but I'll be more explicit in order of decreasing preference:

1. We leave our Ruby as-is
2. We ship Apple's Ruby from https://opensource.apple.com/source/ruby/ruby-131/ruby/
3. We apply patches on top of this existing formula to patch issues that are affecting Homebrew

---

One of the most important things `@mxcl` taught me in the early days of this project was that as a maintainer we should not fix bugs speculatively. What was meant by that is we should not be spending our time fixing issues that are not user reported when there are unclosed and unfixed user reported issues and beyond that not fix issues that are not user visible until all user visible issues are fixed. I think it's worth ruminating on that in the context of this discussion.

JCount:
I respectfully disagree that fixing security issues in a "speculative" manner in Homebrew, which from your above posts would be any time they are not actively being exploited, or explicitly known to be exploitable, to be a waste of time.

I suppose it is a matter of priorities. For me, the existence of widely known security issues, with fixes available, warrants action even if it is not directly known how they might be exploitable in a given use case or do not seem to be particularly dire. This is mainly because I do not personally consider myself to be knowledgable enough, or qualified, to make that call. Hence, I defer to upstream's judgement on the issue when it comes to security fixes, which in the instance included advising all users to update.

However, since it does not seem that we will be making a change, I will drop the matter since the issues do not superficially _seem_ to be critical enough to pick a fight over, and because I have made my opinion on the matter clear.

mistydemeo:
> I've explained my preferences above but I'll be more explicit in order of decreasing preference:

> 1. We leave our Ruby as-is
> 2. We ship Apple's Ruby from https://opensource.apple.com/source/ruby/ruby-131/ruby/
> 3. We apply patches on top of this existing formula to patch issues that are affecting Homebrew

1) is my preference out of the three, so I'll close this PR.

PR #52, opened 2017-11-23
portable-patchelf 0.9 (new formula)

sjackman:
Modify dynamic ELF files.

--------

`patchelf` may be used to relocate ELF files, similarly to `otool` on macOS.
It is required to relocate bottles on Linux.

### DIFF
--- a/Formula/portable-patchelf.rb
+++ b/Formula/portable-patchelf.rb
@@ -1,0 +1,31 @@
+class PortablePatchelf < Formula
+  desc "Modify the dynamic linker and RPATH of ELF executables"
+  homepage "https://nixos.org/patchelf.html"
+
+  url "http://nixos.org/releases/patchelf/patchelf-0.9/patchelf-0.9.tar.gz"
+  sha256 "f2aa40a6148cb3b0ca807a1bf836b081793e55ec9e5540a5356d800132be7e0a"
+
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Please maintain this in a Linuxbrew fork, sorry.

tcurdt:
@MikeMcQuaid Why? This isn't just for linux. I was just looking for the same thing.

sjackman:
I'd be happy to contribute `patchelf` to Homebrew/homebrew-core if it'd desired. It'd be useful to me on macOS for the same reasons that `readelf` in the `binutils` package is useful on macOS, for inspecting ELF files on a macOS system. Our macOS systems at work also have our Linux executables mounted over NFS, and it can be useful to inspect those executables without needing to `ssh` to a Linux server.

sjackman:
Note that I'm planning to submit a PR to Homebrew to implement relocatable bottles (`keg_relocate.rb`) for Homebrew, which requires `patchelf`. Where should that formula live?

MikeMcQuaid:
@tcurdt This tap is just for Homebrew's portable Ruby now.


@sjackman If `patchelf` works on macOS (with the usual new formula constraints) then 👍 to submitting it to Homebrew/homebrew-core.

PR #53, opened 2018-03-09
Tigerbrew portable formulae

mistydemeo:
This adds several formulae which are actively used by Tigerbrew.

The portable curl is used since an increasing number of servers require TLS 1.2 or later, including GitHub. Even ignoring certs, older systems' curls can't fetch from there. After experimenting with a bunch of solutions, including using Tigerbrew's curl package to fetch, I've landed on using a vendored curl with an embedded set of certs as the easiest solution which prevents most forms of user error. This package is also used in the Tigerbrew installer, since most systems that will be installing Tigerbrew don't have a usable git and can't fetch a tarball from GitHub using the system curl.

The expat and git packages are part of an experiment with vendoring a TLS 1.2-compatible git, both for `brew update` and possibly for use in the installer. I've encountered a few issues with users whose systems end up in weird states and whose gits can't switch branches; this may provide a usable solution for that. It may also provide a preferable way to pull down Tigerbrew from GitHub in the installer, instead of using the portable curl.

The commit history is a bit of a mess, but I can clean this up.

### DIFF
--- a/Formula/portable-curl.rb
+++ b/Formula/portable-curl.rb
@@ -1,0 +1,53 @@
+require File.expand_path("../../Abstract/portable-formula", __FILE__)
+
+class PortableCurl < PortableFormula
+  desc "Portable curl"
+  homepage "https://curl.haxx.se/"
+  url "https://curl.haxx.se/download/curl-7.58.0.tar.bz2"
+  mirror "http://curl.askapache.com/download/curl-7.58.0.tar.bz2"
+  sha256 "1cb081f97807c01e3ed747b6e1c9fee7a01cb10048f1cd0b5f56cfe0209de731"
+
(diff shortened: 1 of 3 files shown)


mistydemeo:
The other reason I'm looking at vendoring both curl and git is because I'm preparing for a future in which my bottle domain requires TLS 1.2+ and cannot be fetched from on older Macs. If/when that happens, I'll host the vendored packages on a separate domain under my control and use those to bootstrap the ability to fetch both bottles and source packages.

MikeMcQuaid:
I think these should live in a Tigerbrew fork/tap as when Homebrew needs a custom Git or Curl it uses HTTP downloads for them. Sorry @mistydemeo.

PR #56, opened 2018-04-30
portable-ruby: Add mirrors for the patches

sjackman:
For older versions of curl, work around the error:
```
```

### DIFF
--- a/Formula/portable-ruby.rb
+++ b/Formula/portable-ruby.rb
@@ -30,6 +30,7 @@ class PortableRuby < PortableFormula
   # release we're building.
   patch do
     url "https://github.com/ruby/ruby/commit/b3dbeb6e90f316584f70e33f6bfb9d83fa5f30d3.patch?full_index=1"
+    mirror "https://cdn.rawgit.com/sjackman/1b1a420c39f33512d4ed8129277ba1d2/raw/1cfd510dad86f835092e6ab18528e8cee5e801a2/b3dbeb6e90f316584f70e33f6bfb9d83fa5f30d3.patch"
     sha256 "17a6a37e500f3455bb85e6bd4b077228d7a32f63bf07ecf67248acbd3a5ea724"
   end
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
@sjackman I don't think this is the right solution here. `brew install curl` is already configured to ensure it's built to be able to access GitHub. Does `brew install curl` not fix this?

MikeMcQuaid:
@sjackman May also want to consider adding stripped bottles in here?

sjackman:
```
brew install curl is already configured to ensure it's built to be able to access GitHub. Does brew install curl not fix this?
```
`/usr/bin/curl` on CentOS 5 is not able to access GitHub, since it gives the above error `tlsv1 alert protocol version`. `brew install curl` may fix this issue if it installs `curl` from bottle rather than from source. You prefer that solution?

MikeMcQuaid:
> /usr/bin/curl on CentOS 5 is not able to access GitHub, since it gives the above error tlsv1 alert protocol version

It shouldn't need to access GitHub; only `git` does for the installation Homebrew installation and `brew install curl` should not need to use HTTPS to do so.



> @sjackman May also want to consider adding stripped bottles in here?

To contradict myself: let's leave these as-is; the debug symbols are present on macOS and have been useful in (rare) Homebrew Ruby crashes in the past.

MikeMcQuaid:
> brew portable-package ruby needs HTTPS access to GitHub to download the two patches.

Yes but if `brew install curl` has already been run: why can it not use that `curl` instead?

sjackman:
`brew install curl` has not yet been run. Recall: `brew install curl` may fix this issue if it installs curl from bottle rather than from source. You prefer that solution?

sjackman:
I was avoiding installing any brewed packages to avoid any possible interactions between the `foo` and `portable-foo` packages. My caution may not be warranted, if the `portable-foo` packages are sufficiently isolated from the `foo` packages.

MikeMcQuaid:
> My caution may not be warranted, if the portable-foo packages are sufficiently isolated from the foo packages.

They are sufficiently isolated, `brew portable-package` and CI test this isolation and this is the same workflow (`brew install curl`) as required on Leopard to build the Ruby there.

sjackman:
Feel free to close this issue.

MikeMcQuaid:
Cool, thanks @sjackman.

PR #61, opened 2018-05-02
portable-berkeley-db (new formula)

sjackman:
portable-ruby: Depends on portable-berkeley-db for Linux

Please do not merge until I've had a chance to test the resulting bottle.


sjackman:
```
/home/linuxbrew/.linuxbrew/Cellar/portable-ruby/2.3.3_2/bin/ruby -rdbm -e "DBM.new('test')"
echo $?
0
```
🍏

Formula/portable-berkeley-db.rb
```diff
@@ -0,0 +1,78 @@
...
+  url "https://download.oracle.com/berkeley-db/db-6.2.32.tar.gz"
+  sha256 "a9c5e2b004a5777aa03510cfe5cd766a4a3b777713406b02809c17c8e0e7a8fb"
+
+  bottle do
+    cellar :any
+    sha256 "ee2f38644137df02f3ebd9fa689a09f1e12c2a21dab38a03f77e2c213a6135ef" => :high_sierra
+    sha256 "eb54f8ab1d0149e073f641ad066e6ffb179afdee83cd3211e90eeaaaa4a7bc9a" => :sierra
+    sha256 "c084857cfdd9bbc5eec028cc551d2323c050489e8a963ec40076270db8d14fb3" => :el_capitan
+    sha256 "1ed9a471cd8a38b58053a86e37b4ec83c09944781158fa20e2ba300f3b374f4f" => :yosemite
+  end
```
xu-cheng: You should remove the bottle block.

Formula/portable-berkeley-db.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  depends_on :java => [:optional, :build]
```
xu-cheng: We can remove the optional deps in the bottle.

Formula/portable-berkeley-db.rb
sjackman: Woops. Thanks.

PR #65, opened 2018-07-13
Test and fix Dockerfiles

MikeMcQuaid:
- Build x86_64 Dockerfile on Travis CI (ARM one is broken)
- Simplify and update README instructions
- Use Debian Wheezy for both Dockerfiles
- Make Dockerfiles more consistent

### DIFF
--- a/docker/Dockerfile.x86_64
+++ b/docker/Dockerfile.x86_64
@@ -1,40 +1,33 @@
 # vim: set ft=dockerfile:
 
 # We want a sufficiently old glibc to be backwards compatible.
-FROM centos:centos5
-MAINTAINER Homebrew Maintainers <homebrew-discuss@googlegroups.com>
+FROM debian:wheezy
 
-# CentOS 5 has reached end-of-life, as result all of mirrors have stopped working.
-# Switch to use http://vault.centos.org/ instead.
-RUN sed -i '/mirrorlist=.*/d' /etc/yum.repos.d/{CentOS-Base,libselinux}.repo && \
-    sed -i 's|#baseurl=http://mirror.centos.org/centos/$releasever|baseurl=http://vault.centos.org/5.11|g' /etc/yum.repos.d/{CentOS-Base,libselinux}.repo
-
-RUN rpm -Uvh http://dl.fedoraproject.org/pub/archive/epel/5/x86_64/epel-release-5-4.noarch.rpm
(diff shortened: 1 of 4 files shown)


MikeMcQuaid:
@sjackman FYI this was the Dockerfile I used to build the x86_64 container. The other Dockerfile would not build correctly. I'd suggest that once the ARM container is fixed we have the Linux CI here just building inside Docker. Would you be able to pick that up?

sjackman:
Yes, once I'm back from vacation, I can look into that.

docker/Dockerfile.x86_64
```diff
@@ -1,40 +1,33 @@
...
     HOMEBREW_BUILD_BOTTLE=1 \
     HOMEBREW_BUILD_FROM_SOURCE=1 \
-    HOMEBREW_CC=cc \
     HOMEBREW_FORCE_VENDOR_RUBY=1 \
     HOMEBREW_NO_ANALYTICS=1 \
     HOMEBREW_NO_AUTO_UPDATE=1
 
-RUN echo 'insecure' >> /home/linuxbrew/.curlrc
 RUN git clone --depth=1 https://github.com/Linuxbrew/brew /home/linuxbrew/.linuxbrew && \
     brew analytics off && \
```
zbeekman: It looks like linuxbrew is installing portable-ruby 2.3.3_2 when running any `brew` command. This install is failing with:

```
portable-ruby/2.3.3_2/README.md
+ safe_cd /home/linuxbrew/.linuxbrew/Library/Homebrew/vendor/portable-ruby
+ cd /home/linuxbrew/.linuxbrew/Library/Homebrew/vendor/portable-ruby
+ quiet_stderr ./2.3.7/bin/ruby --version
+ [[ -z '' ]]
+ command ./2.3.7/bin/ruby --version
+ rm -rf 2.3.7
+ [[ -d 2.3.7.reinstall ]]
+ odie 'Failed to install vendor ruby.'
+ onoe 'Failed to install vendor ruby.'
+ [[ -t 2 ]]
+ echo -n 'Error: '
Error: + [[ 1 -eq 0 ]]
+ echo 'Failed to install vendor ruby.'
Failed to install vendor ruby.
+ exit 1
Error: Failed to install vendor Ruby.
```

zbeekman: (but then it's expecting 2.3.7 to be present and dying when it can't execute 2.3.7)

PR #66, opened 2018-10-06
.travis.yml: Run brew portable-package in Docker

sjackman:


MikeMcQuaid:
Now CI is 💚 to do this: 👍 👏

PR #72, opened 2019-01-23
Deprecate pre-Mavericks.

MikeMcQuaid:
This will be used to build a new Portable Ruby for Homebrew 2.0.0 on Mavericks.

Should not be merged until Homebrew 2.0.0 will be released imminently.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,9 +1,11 @@
 jobs:
 - job: macOS
   pool:
-    vmImage: xcode9-macos10.13
+    vmImage: macOS-10.13
   steps:
     - bash: |
+        set -e
+        sudo xcode-select --switch /Applications/Xcode_10.1.app/Contents/Developer
         export HOMEBREW_DEVELOPER=1
         export HOMEBREW_NO_AUTO_UPDATE=1
         brew update-reset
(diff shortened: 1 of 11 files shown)


azure-pipelines.yml
```diff
@@ -28,5 +28,6 @@ jobs:
         PATH="$HOME/.linuxbrew/bin:$HOME/.linuxbrew/sbin:$PATH"
         ln -s "$PWD" "$HOME/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-portable-ruby"
         docker build -f docker/Dockerfile.x86_64 -t homebrew-portable-x86_64 .
+        # TODO: this isn't running the same version as Azure Pipelines have checked out.
```
MikeMcQuaid: @sjackman I need an assist here: the docker image run isn't quite correct as it's tapping the `master` branch and not the checked out branch in Azure Pipelines (e.g. from the branch, PR or fork). Mind taking a look?

MikeMcQuaid:
@sjackman Yes, that's related to my comments above. That code has been removed in this PR but isn't being used in the Docker image because it's tapping this tap without changing the revision.

azure-pipelines.yml
sjackman: I'm happy to. Are you able to wait a few days? I'm on a thesis deadline crunch.

Here's the line of code that taps `Homebrew/portable-ruby`:
https://github.com/Homebrew/homebrew-portable-ruby/blob/2cce4a10d4b0d24585fa9d14a897cd1bffa9273a/docker/Dockerfile.x86_64#L25

I'd suggest not tapping it in the Dockerfile, so that you can instead  tap the appropriate version in `azure-pipelines.yml` after `docker build` and before `brew portable-package`.
https://github.com/Homebrew/homebrew-portable-ruby/blob/2cce4a10d4b0d24585fa9d14a897cd1bffa9273a/azure-pipelines.yml#L30-L31

sjackman:
That was a just a note to self. I like to copy-and-paste the error message from the build log into the issue, so I see it right away when I revisit the issue.

MikeMcQuaid:
> That was a just a note to self. I like to copy-and-paste the error message from the build log into the issue, so I see it right away when I revisit the issue.

Gotcha 👍

azure-pipelines.yml
MikeMcQuaid: @sjackman Good idea, I'll give that a go.

sjackman: Alternatively, a `docker cp` command could copy the Homebrew/portable-ruby files from Azure to the Docker container.

MikeMcQuaid:
@sjackman Have removed the failing `docker build` for now and it can be readded later when this is working.

sjackman:
Okay. Thanks for the heads up.

PR #73, opened 2019-02-22
Add docker/Dockerfile.arm64

sjackman:


MikeMcQuaid:
I presume this can't be run in CI. Is there any way to e.g. have the syntax linted somehow e.g. run some `docker` command that prints metadata that will error if the Dockerfile doesn't work?

docker/Dockerfile.arm64
```diff
@@ -0,0 +1,21 @@
...
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+        ca-certificates curl file g++ git-core locales make patch \
+    && rm -rf /var/lib/apt/lists/* \
+    && localedef -i en_US -f UTF-8 en_US.UTF-8 \
+    && useradd -m -s /bin/bash linuxbrew \
+    && git clone --depth=1 https://github.com/Homebrew/brew /home/linuxbrew/.linuxbrew/Homebrew \
+    && mkdir /home/linuxbrew/.linuxbrew/bin \
+    && ln -s ../Homebrew/bin/brew /home/linuxbrew/.linuxbrew/bin/ \
+    && brew tap homebrew/portable-ruby
```
zbeekman: Don't you need to change users to linuxbrew? Won't all the `/home/linuxbrew` dirs be owned by root?

Also, I think you need to tap `homebrew/core` too. Also, the other docker files run `brew analytics off` explicitly. And install patchelf, etc.

sjackman: > Don't you need to change users to linuxbrew? Won't all the `/home/linuxbrew` dirs be owned by root?

We only care about what ends up in the bottle tarball. The files in the tarball will owned by `root:root` rather than `linuxbrew:linuxbrew` as we've been doing until now. So that is a change. Normally the owner of the file stored in the tarball is ignored when you extract the tarball.

sjackman: > Also, I think you need to tap `homebrew/core` too.

That's done automatically.

sjackman: > Also, the other docker files run `brew analytics off` explicitly.

We have `ENV HOMEBREW_NO_ANALYTICS=1`

sjackman:
> I presume this can't be run in CI. Is there any way to e.g. have the syntax linted somehow e.g. run some `docker` command that prints metadata that will error if the Dockerfile doesn't work?

Not easily. It is possible inside a QEMU emulator with some work.

We can run either `dockerfile_lint` or `dockerfilelint` to lint the syntax. I'll add that.

sjackman:
Azure on Linux
```
Step 1/15 : FROM debian:wheezy
unauthorized: authentication required
```
https://dev.azure.com/Homebrew/Homebrew/_build/results?buildId=4503&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&taskId=8ef82b3b-1feb-5bbd-06f6-b1f7b5467f03&lineStart=14&lineEnd=15

That's… weird.

sjackman:
The `docker pull` error was transient. It succeeded this time around.

zbeekman:
@MikeMcQuaid

> I presume this can't be run in CI. Is there any way to e.g. have the syntax linted somehow e.g. run some `docker` command that prints metadata that will error if the Dockerfile doesn't work?

If there's a way to connect east-arm-01.izk.io into the azure pipeline, then we could test it there. I can create an azure user, or similar, but not sure where to start.

docker/Dockerfile.arm64
zbeekman: > We only care about what ends up in the bottle tarball. The files in the tarball will owned by `root:root` rather than `linuxbrew:linuxbrew` as we've been doing until now. So that is a change. Normally the owner of the file stored in the tarball is ignored when you extract the tarball.

We could always just `chown -R linuxbrew:linuxbrew` and add the linuxbrew user back... it's worth testing that un-priveledged users can build etc., IMO.

sjackman: The `linuxbrew` user is still created. It's just not the current user.

> it's worth testing that un-priveledged users can build etc., IMO.

You mean that un-privileged users can run `brew portable-package` to build portable bottles?

zbeekman: Yes... but maybe that's dumb. As long as poured bottle perms get sorted out correctly, it probably doesn't matter.

PR #75, opened 2019-02-25
portable-openssl: Fix for ARM64 on Linux

sjackman:


Formula/portable-openssl.rb
```diff
@@ -25,11 +25,15 @@ def openssldir
...
     if OS.mac?
-      %w[darwin64-x86_64-cc enable-ec_nistp_64_gcc_128]
-    elsif Hardware::CPU.arm?
-      %w[linux-armv4]
+      ["darwin64-x86_64-cc", "enable-ec_nistp_64_gcc_128"]
     else
-      %w[linux-x86_64]
+      args = ["enable-md2"]
+      if Hardware::CPU.intel?
+        args << (Hardware::CPU.is_64_bit? ? "linux-x86_64" : "linux-elf")
```
MikeMcQuaid: Think a nested `if` would be preferable/more readable here.

PR #77, opened 2019-02-25
Azure: Lint the Dockerfiles

sjackman:
Fix style of Dockerfile.arm and Dockerfile.x86_64

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+package-lock.json
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,6 +19,11 @@ jobs:
   steps:
     - bash: |
         set -e
+        sudo npm install -g dockerfilelint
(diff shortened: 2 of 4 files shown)


sjackman:
I first tried `npm install dockerfilelint`, which locally creates the executable file `node_modules/.bin/dockerfilelint`. On Azure, after `npm install dockerfilelint`, which appears successful, no `node_modules` directory is created. Is that because it's being run as root? Where do the local modules get installed, if not into the `node_modules` in the current working directory?

sjackman:
```console
$ dockerfilelint docker/Dockerfile.*
File:   docker/Dockerfile.arm
Issues: None found 👍

File:   docker/Dockerfile.arm64
Issues: None found 👍

File:   docker/Dockerfile.x86_64
Issues: None found 👍
```
🎆

MikeMcQuaid:
Thanks @sjackman!

PR #89, opened 2019-09-26
portable-package: Install dependencies from source

### LINKED ISSUE #87: Run brew portable-package on Linux CI
See:
https://github.com/Homebrew/homebrew-portable-ruby/blob/4fe159fc08e2dffb17146c6938ce162075d60fcb/.github/workflows/tests.yml#L40-L43

Currently this doesn't do anything but it should be configured to use this repository as expected.

sjackman:
Installing bottled dependencies requires brewed glibc.

portable-libyaml: Avoid autotools dependency

Fixes #87

### DIFF
--- a/cmd/brew-portable-package.rb
+++ b/cmd/brew-portable-package.rb
@@ -25,12 +25,14 @@ BOTTLE_ARGS = %w[
 ARGV.named.each do |name|
   name = "portable-#{name}" unless name.start_with? "portable-"
   begin
+    deps = Utils.popen_read("brew", "deps", "--include-build", name).split("\n")
+
+    # Avoid installing glibc. Bottles depend on glibc.
+    safe_system "brew", "install", "-s", *deps if OS.linux?
+
     safe_system "brew", "install", "--build-bottle", name
     unless ARGV.include? "--no-uninstall-deps"
-      deps = Utils.popen_read("brew", "deps", "--include-build", name).split("\n")
-      deps.each do |dep|
-        safe_system "brew", "uninstall", "--force", "--ignore-dependencies", dep
-      end
+      safe_system "brew", "uninstall", "--force", "--ignore-dependencies", *deps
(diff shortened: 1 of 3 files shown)


cmd/brew-portable-package.rb
```diff
@@ -25,12 +25,11 @@
 ARGV.named.each do |name|
   name = "portable-#{name}" unless name.start_with? "portable-"
   begin
+    deps = Utils.popen_read("brew", "deps", "--include-build", name).split("\n")
+    safe_system "brew", "install", "-s", *deps
```
MikeMcQuaid: This should be guarded to Linux-only I think.

MikeMcQuaid:
Looks like still linking against `glibc` unfortunately: https://github.com/Homebrew/homebrew-portable-ruby/pull/89/checks?check_run_id=238026471

cmd/brew-portable-package.rb
sjackman: Done.

xu-cheng:
FYI, if you have trouble to use system gcc, you should probably set `HOMEBREW_CC` in the docker.

https://github.com/Homebrew/homebrew-portable-ruby/commit/fb53f057d82c2590e63ef6240f7a8e4aa5ed9b39

You may also want to set `HOMEBREW_BUILD_FROM_SOURCE`.

I think these env were removed incorrectly.

MikeMcQuaid:
> You may also want to set `HOMEBREW_BUILD_FROM_SOURCE`.

This variable no longer works.

sjackman:
Fingers crossed, this PR should work after PR https://github.com/Homebrew/brew/pull/6505 is merged.

MikeMcQuaid:
CI is 💚, nice work @sjackman and thanks again!

sjackman:
Thanks for merging, Mike!

PR #90, opened 2019-10-09
CI: Install and test portable-ruby [WIP]

sjackman:
Install and test the portable-ruby bottle.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,7 +37,28 @@ jobs:
     - name: Build Portable Ruby
       run: |
         if [ "$RUNNER_OS" = "Linux" ]; then
-          docker run -v"$PWD":/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-portable-ruby homebrew-portable-x86_64 brew portable-package portable-ruby
+          docker run --name=homebrew-portable-ruby -w /bottle -v"$PWD":/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-portable-ruby homebrew-portable-x86_64 brew portable-package portable-ruby
+          docker cp homebrew-portable-ruby:/bottle .
+          mv bottle/* .
+          rmdir bottle
         else
           brew portable-package portable-ruby
         fi
+
+    - name: Install Portable Ruby
+      run: |
+        version=2.6.3
+        if [ "$RUNNER_OS" = "Linux" ]; then
+          prefix=/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/vendor
+        else
+          prefix=/usr/local/Homebrew/Library/Homebrew/vendor
+        fi
+        mkdir -p $prefix
+        tar -C $prefix -xf portable-ruby--*.tar.gz
+        ln -sf $version $prefix/portable-ruby/current
+        echo $version >$prefix/portable-ruby-version
+
+    - name: Test Portable Ruby
+      run: |
+        brew config
+        brew tests


.github/workflows/tests.yml
```diff
@@ -37,7 +37,28 @@ jobs:
...
+          rmdir bottle
         else
           brew portable-package portable-ruby
         fi
+
+    - name: Install Portable Ruby
+      run: |
+        version=2.6.3
+        if [ "$RUNNER_OS" = "Linux" ]; then
+          prefix=/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/vendor
```
MikeMcQuaid: Might need to do this in the docker container on Linux?

sjackman:
GitHub CI on Linux
```
mkdir: cannot create directory ‘/home/linuxbrew’: Permission denied
```
https://github.com/Homebrew/homebrew-portable-ruby/pull/90/checks?check_run_id=253953687#step:8:13

MikeMcQuaid:
@sjackman You'll need `sudo`

.github/workflows/tests.yml
```diff
@@ -37,7 +37,28 @@ jobs:
...
+        fi
+        mkdir -p $prefix
+        tar -C $prefix -xf portable-ruby--*.tar.gz
+        ln -sf $version $prefix/portable-ruby/current
+        echo $version >$prefix/portable-ruby-version
+
+    - name: Test Portable Ruby
+      run: |
+        brew config
+        brew tests
```
MikeMcQuaid: The main issue I see with the `brew tests` here is: what do you do if it fails? In the case of https://github.com/Homebrew/brew/pull/6556 there were some tests (extracted to https://github.com/Homebrew/brew/pull/6558) that could be fixed before 2.6.3 was merged and worked on 2.6.3 and 2.3.7. Unfortunately, though, there were some tests (https://github.com/Homebrew/brew/pull/6556/commits/e3ee64a5fd59f4f3203b369424bf8347b563cb11) which couldn't be fixed to work consistently on both versions. That would mean that neither this repo nor Homebrew/brew can have green CI until both have been updated at once (or the bottle is built from a WIP branch here, shipped to `master` on Homebrew/brew and then the CI rerun and hope it passes).

xu-cheng:
Some suggestions if you are interested.

* After setting `HOMEBREW_DEVELOPER`, you can set `HOMEBREW_RUBY_PATH` directly to use that ruby.
* For Linux, you may want to test the portable-ruby on several Linux distros  using docker. The choice of distros can be based on your analytical data. So you can do something like this:
```
tar -xzvf /path/to/bottle/file
git clone --depth=1 https://github.com/Homebrew/brew
cat > test.env <<EOS
HOMEBREW_DEVELOPER=1
HOMEBREW_RUBY_PATH=/test/portable-ruby/<version>/bin/ruby
PATH=/test/brew/bin:$PATH
EOS
docker run --rm --env-file test.env -v $(pwd):/test/ debian:latest brew tests
docker run --rm --env-file test.env -v $(pwd):/test/ centos:latest brew tests
```
* You could also add ARM build and test in CI. See https://github.com/Homebrew/brew/pull/6595#discussion_r334750572

.github/workflows/tests.yml
xu-cheng: In this case, you could temporarily change brew clone url to a forked in CI here. So you can test both portable-ruby and brew at the same time.

MikeMcQuaid:
Good ideas, thanks @xu-cheng.

sjackman:
@xu-cheng Is this PR superseded by PR https://github.com/Homebrew/homebrew-portable-ruby/pull/90?

zbeekman:
> @xu-cheng Is this PR superseded by PR #90?

I think so.

PR #91, opened 2019-10-15
ci: build and test portable-ruby

xu-cheng:


docker/Dockerfile
```diff
@@ -0,0 +1,33 @@
+ARG arch=amd64
+FROM ${arch}/ruby:2.6.3-stretch
```
xu-cheng: This seems to be a problem. The glibc in Debian stretch is too new. And there is no old Debian available for official ruby dockers.

I will explore whether it is possible to use alpine and musl-libc to build the Ruby.

MikeMcQuaid: Can `FROM debian/eol:wheezy` not be used? Why does it need to be a Ruby docker?

xu-cheng: > Can FROM debian/eol:wheezy not be used?

I will take a look. It seems that `debian/eol:wheezy` indeed supports multiarch. However, I am not sure how to pull docker image for a specific arch as they all share the same tag. It seems that I have to pull using the hash of the image.

> Why does it need to be a Ruby docker?

Because we need it to run Homebrew. This is specially important if there is not portable-ruby available. For example, currently we need an external ruby to run Homebrew on ARM as there is no portable-ruby 2.6.3 for ARM. Basically, this is a bootstrap problem. I was thinking to use Ruby docker to avoid it.

MikeMcQuaid: > I will take a look. It seems that `debian/eol:wheezy` indeed supports multiarch. However, I am not sure how to pull docker image for a specific arch as they all share the same tag. It seems that I have to pull using the hash of the image.

I reckon use that image as-is for Linux and we can deviate for the ARM builds.

> Basically, this is a bootstrap problem. I was thinking to use Ruby docker to avoid it.

👍 seems like a good idea for ARM and we could consider for x86_64 if it's a problem after that?

README.md
```diff
@@ -16,9 +16,9 @@ Run `brew portable-package ruby` inside an OS X 10.9 VM (so it is compatible wit
 
 Build a Docker image for your architecture by running one of the following commands.
 
-- `docker build -f docker/Dockerfile.x86_64 -t homebrew-portable .`
-- `docker build -f docker/Dockerfile.arm -t homebrew-portable .`
-- `docker build -f docker/Dockerfile.arm64 -t homebrew-portable .`
+- `docker build -f docker/Dockerfile --platform linux/amd64 --build-arg os=wheezy -t homebrew-portable .`
+- `docker build -f docker/Dockerfile --platform linux/arm/v7 --build-arg os=wheezy -t homebrew-portable .`
+- `docker build -f docker/Dockerfile --platform linux/arm64/v8 --build-arg os=jessie -t homebrew-portable .`
```
sjackman: Please move `docker/Dockerfile` to `./Dockerfile`

sjackman:
Thanks for taking up this torch, @xu-cheng !

xu-cheng:
So FYI, here is the current state of this PR.

* The CI can now successfully build and test portable ruby on macOS and Linux x86_64.
* For ARM64v8, the building should be fine. I need to fix its test though by using a different docker image. I will take a look at this issue late this week.
* One outstanding issue is that running ARM binaries using QEMU simulation  is extremely slow. As such, it takes several hours to build the portable ruby. As far as I am aware of, there are two solutions to improve the performance in the future: (1) using native hardware; (2) using cross-compiling.
* BTW, the current portable ruby for ARM64v8 merged in https://github.com/Homebrew/brew/pull/6595 should be updated. I just realized it was built against the glibc which is too new.  I have built a new one in CI with old glibc. It can be downloaded from https://github.com/Homebrew/homebrew-portable-ruby/suites/267549269/artifacts/132613 Though, test should be done before uploading it to Homebrew.
* I was not able to make portable ruby for ARM32v6 work for now. The docker does provide official images for both ARM32v6 and ARM32v7. However, I have encountered following problems:
   * The old portable ruby (https://github.com/Homebrew/homebrew-portable-ruby/releases/download/2.3.7/portable-ruby-2.3.7.armv6_linux.bottle.tar.gz) cannot be used in ARM32v6. The reason is that the old portable ruby requires dynamic linker at `/lib/ld-linux-armhf.so.3` whereas the docker image provides the dynamic linker at `/lib/ld-linux.so.3`. I am not sure whether this is the problem of the docker image or the old portable ruby. If someone has the actual hardware with ARM32v6 CPU, it would be helpful to let me know which dynamic linker should I build the portable ruby against. If it is `/lib/ld-linux-armhf.so.3`, I cannot find suitable docker image.
   * For ARM32v7, the dynamic linker is `/lib/ld-linux-armhf.so.3`.  So I can use the old portable ruby to bootstrap. However, we need to fix a bug in Homebrew to build ARM32v7 binaries. Currently, it passes the wrong `-march` flag.
   * All in all, I am not sure which arch we should support ARM32v6 or ARM32v7. It does not seem we can use the portable-ruby for them as different dynamic linkers are used.

/cc @sjackman for comments.

docker/Dockerfile
sjackman: > I will take a look. It seems that debian/eol:wheezy indeed supports multiarch. However, I am not sure how to pull docker image for a specific arch as they all share the same tag. It seems that I have to pull using the hash of the image.

Does not it not pull the appropriate image for the current architecture by default?

MikeMcQuaid:
Thanks @xu-cheng. I'd like to propose (to @sjackman) that given the above we don't try to support both ARM versions for portable Ruby or, ideally, either. I don't see ARM machines as being nearly as likely for the end-user to not have root and we could instead modify the `vendor-install` code to instruct those users to install Ruby 2.6 manually through their distro instead.

cmd/brew-portable-package.rb
```diff
@@ -12,6 +12,7 @@
 #:    A new commit will then be generated unless `--no-commit` is passed.
 
 ENV["HOMEBREW_DEVELOPER"] = "1"
+ENV["HOMEBREW_BUILD_BOTTLE"] = "1" if OS.linux?
```
MikeMcQuaid: Why is this needed? If it's for `safe_system "brew", "install", "-s", *deps` can you change that to `--build-bottle` instead?

xu-cheng: Coming to think about it, I think all portable formulae should be built with `--build-bottle` flag. This is required to ensure portable formulae are built with correct `-march` flag. Moreover, `--build-bottle` or `HOMEBREW_BUILD_BOTTLE` only affect current formula instead of its dependencies.

So I think we should change it to something like this:
```ruby
    deps = Utils.popen_read("brew", "deps", "--include-build", name).split("\n")

    graph = {}
    deps.each { |dep| graph[dep] = Formula[dep].deps.map(&:name) }
    deps = TSort.tsort ->(&b) { graph.each_key(&b) }, ->(n, &b) { graph[n].each(&b) }

    portable_deps = deps.select { |dep| dep.start_with? "portable-" }
    other_deps = deps - portable_deps

    # Avoid installing glibc. Bottles depend on glibc.
    other_deps.each { |dep| safe_system "brew", "install", "-s", dep } if OS.linux?

    portable_deps.each { |dep| safe_system "brew", "install", "--build-bottle", dep }
    safe_system "brew", "install", "--build-bottle", name
```

If it's OK to you, I can send a separated PR for this.

MikeMcQuaid: `--build-bottle` will use the same `march` on macOS, FYI. The main issue is that it will never run `post_install` (which shouldn't be an issue here but is worth being explicit).

I don't think it's necessary to do more than just use `--build-bottle` instead of `-s`. We don't need to build e.g. `pkg-config` as a bottle.

xu-cheng: > --build-bottle will use the same march on macOS, FYI.

I don't think this is true for linux.

> I don't think it's necessary to do more than just use --build-bottle instead of -s. We don't need to build e.g. pkg-config as a bottle.

In the above code, `--build-bottle` is only passed to portable formulae.

MikeMcQuaid: > In the above code, `--build-bottle` is only passed to portable formulae.

If the others are being built from source anyway (on Linux): I don't think it matters. I'm open to changing the existing `-s` to `--build-bottle` on Linux but I don't see a need in going beyond that.

Let's not make changes that aren't demonstrated to be necessary through e.g. reproducible bugs.

xu-cheng: > Let's not make changes that aren't demonstrated to be necessary through e.g. reproducible bugs.

The reason that there is no bug for now is because we currently set `HOMEBREW_BUILD_BOTTLE` in the docker. This is effectively equivalent to passing `--build-bottle` to all formulae on Linux. If we remove `HOMEBREW_BUILD_BOTTLE`, the portable ruby won't work on old CPU.

> If the others are being built from source anyway (on Linux): I don't think it matters. I'm open to changing the existing -s to --build-bottle on Linux but I don't see a need in going beyond that.

This is not enough. `--build-bottle` flag only affects the current formula not its deps.

MikeMcQuaid: Ok, leave this as-is.

sjackman:
> Thanks @xu-cheng. I'd like to propose (to @sjackman) that given the above we don't try to support both ARM versions for portable Ruby or, ideally, either. I don't see ARM machines as being nearly as likely for the end-user to not have root and we could instead modify the vendor-install code to instruct those users to install Ruby 2.6 manually through their distro instead.

Building the `portable-ruby` bottle for ARM32 is quite a bit slower, taking hours, whether on real AMR hardware or emulating. I'm okay with dropping ARM32.

Building the `portable-ruby` bottle for AArch64 using @zbeekman's ARM64 hardware was quite painless, and took half an hour. I suggest that we build `portable-ruby` for `x86_64` on macOS and Linux, but not for ARMv6 and ARM64, just for reasons of speed. Given how rarely we update `portable-ruby`, building the bottle manually doesn't seem like a big issue to me. We can include instructions in the `README.md` how to build the bottle for ARM64 using either real hardware or emulating using Docker Desktop for macOS.

sjackman:
> All in all, I am not sure which arch we should support, ARM32v6 or ARM32v7? It does not seem we can use the same portable-ruby for both of them as different dynamic linkers are used.

It may be moot, but use the dynamic linker used by Raspbian, which is `/lib/ld-linux-armhf.so.3`. The primary reason to support ARM32 was for Raspberry Pi, which is a 32-bit user space, even when the hardware is 64-bit, like with the Raspberry Pi 4. I wonder if building `portable-ruby` on the Raspberry Pi 4 would be faster.

sjackman:
> May I ask what hardware is Raspbian used? ARM32v6 or ARM32v7.

Raspbian the Linux distribution is targeted for ARMv6 to support running on the Raspberry Pi 1. They have only a single distribution that runs on all versions of the Raspberry Pi hardware.

Raspberry Pi the hardware:
RPi1 ARMv6
RPi2 ARMv7
RPi3 ARMv8, which supports both 64-bit AArch64 and 32-bit ARMv8
RPi4 ARMv8, which supports both 64-bit AArch64 and 32-bit ARMv8
See https://en.wikipedia.org/wiki/Raspberry_Pi#Specifications

>  FYI, I can make ARM32v7 portable-ruby work with some bug fixes on Homebrew side.

What bug fixes are needed? Are far as I know, Homebrew/brew works fine on 32-bit ARM hardware, unless there's been bit rot.

sjackman:
I wonder whether the latest version of Raspbian has Ruby 2.6.3 or newer.

xu-cheng:
I think this PR is almost finished. Here are my proposal for moving forward.

*  The current ARM64 portable-ruby should be replaced as it is against glibc too new. We should update it to the one built in this PR. Since I don't have access to the release or Bintray, @sjackman would you mind to help this? You can download the new bottle here: https://github.com/Homebrew/homebrew-portable-ruby/suites/268830530/artifacts/136186
* I will send a PR to homebrew to fix `brew tests` for ARM. The current test failed at some small problems, which should be easy to fix.
* I will drop the support for AMR32 for now. But I can re-add support ARM32v7 (not ARM32v6) in the future.
* There is a small problem. The docker linter reports a problem, which I believe is a false alarm. Should we remove the docker linter test in this case?
* All in all, I think this PR is almost ready to be merged (after the portable-ruby for ARM64 is uploaded to Homebrew).

xu-cheng:
> What bug fixes are needed? Are far as I know, Homebrew/brew works fine on 32-bit ARM hardware, unless there's been bit rot.

At least from my test using docker and simulation, the bottle buildings are failed due to incorrect `-march=armv6` flag. If we pass `-march=armv7`, it will work fine.

sjackman:
>> What bug fixes are needed? Are far as I know, Homebrew/brew works fine on 32-bit ARM hardware, unless there's been bit rot.

> At least from my test using docker and simulation, the bottle buildings are failed due to incorrect -march=armv6 flag. If we pass -march=armv7, it will work fine.

It's possible and fine to target `-march=armv6` on `armv7` hardware. We do a similar thing on `x86_64` where we target the oldest supported macOS Intel hardware, regardless of which hardware we're actually building on.

That said, I'm fine dropping support for `armv6` and moving to `armv7`, which drops support for the Raspberry Pi 1.

sjackman:
> There is a small problem. The docker linter reports a problem, which I believe is a false alarm. Should we remove the docker linter test in this case?

Yeah, false positive. I'm fine with disabling just that one test.

sjackman:
> The current ARM64 portable-ruby should be replaced as it is against glibc too new.

I can do that later today.

zbeekman:
I think dropping armv6 is fine. rPi 1 is pretty old at this point.

I like the idea of cross compiling for ARM to speed things up, but given my (admittedly somewhat superficial) understanding of the complexities of homebrew compiler selection and environment sanitation makes me think that cross compiling portable ruby in a chroot will be challenging.

sjackman:
Cross compiling is likely to be quite difficult to get working in Homebrew.

MikeMcQuaid:
> Building the `portable-ruby` bottle for ARM32 is quite a bit slower, taking hours, whether on real AMR hardware or emulating. I'm okay with dropping ARM32.

👍 

> Building the `portable-ruby` bottle for AArch64 using @zbeekman's ARM64 hardware was quite painless, and took half an hour. I suggest that we build `portable-ruby` for `x86_64` on macOS and Linux, but not for ARMv6 and ARM64, just for reasons of speed. Given how rarely we update `portable-ruby`, building the bottle manually doesn't seem like a big issue to me.

As the person who is rushing to get `portable-ruby` changes out based on Apple's unpredictable changes: it's a big deal to me. I'd rather we have something that I can run in a local Dockerfile (and in GitHub Actions) that might take several hours over something that relies on having access to the infrastructure of another maintainer.

---

The big questions that I'd love to be answered before we make any decisions on what we support/don't:
- do we have any analytics on how many users there are of any of these bottles? If not: what are the download numbers like?
- do any of the common Linux distributions for ARM provide Ruby 2.6.3? If not: what's the newest version of Ruby they provide? Are there any back ports online for a newer Ruby?

sjackman:
> As the person who is rushing to get portable-ruby changes out based on Apple's unpredictable changes: it's a big deal to me. I'd rather we have something that I can run in a local Dockerfile (and in GitHub Actions) that might take several hours over something that relies on having access to the infrastructure of another maintainer.

CI needs to build and test `portable-ruby` for x86_64 on macOS and Linux using GitHub Actions and Docker. Those are our primary supported targets. Get those merged and shipped together.

In a second step, a developer (like me or Zaak) can build a bottle for AArch64 and/or ARM32 using emulation in Docker or real ARM hardware. That way ARM doesn't hold up our primary targets.

sjackman:
> do any of the common Linux distributions for ARM provide Ruby 2.6.3? If not: what's the newest version of Ruby they provide? Are there any back ports online for a newer Ruby?

| Distribution | Ruby Version |
|-----|-----|
| Ubuntu 16.04 (Xenial Xerus) | 2.3.1 |
| Ubuntu 18.04 (Bionic Beaver) | 2.5.1 |
| Debian 10 | 2.5.5 |

xu-cheng:
I have now find a way to build ARM32v6 portable ruby. Now I need help to upload them to releases and Bintray. @sjackman 

* ARM64v8: https://github.com/Homebrew/homebrew-portable-ruby/suites/268830530/artifacts/136186
* ARM32v6: https://github.com/xu-cheng/homebrew-portable-ruby/suites/271156010/artifacts/142706

After this, I will be able to finalize this PR so it can be ready to be merged.

xu-cheng:
This PR now only depends on: 
* [ ] Final fix on `brew tests`:https://github.com/Homebrew/brew/pull/6618
* [ ] Upload ARM32v6 and ARM64v8 bottles to:
  * [ ] Release
  * [ ] Linuxbrew Bintray
* [ ] Update`vendor-install`

zbeekman:
If anyone else needs access to a physical aarch64 let me know. I think I can get it configured to run docker, so that would allow native container workflows. I sit on the same power grid as the White House (I think---it's nearby) and I've never had a power outage in the ~10 years that I've lived here. I'd be happy to maintain the hardware and make it available to Homebrew & maintainers if it would be useful. I agree with @MikeMcQuaid that we don't want to rely on maintainers' private hardware, but if it can help bootstrap docker images because of the native hardware, then I'm happy to lend it to the cause.

The machine in question is an RK3399 RockPro64 with a WD black NVMe SSD mounted via PCIe. I'm working on getting docker sorted as we speak.

MikeMcQuaid:
> In a second step, a developer (like me or Zaak) can build a bottle for AArch64 and/or ARM32 using emulation in Docker or real ARM hardware. That way ARM doesn't hold up our primary targets.

The problem with that approach is if you're bumping portable Ruby then anyone relying on the portable Ruby on ARM will have their Homebrew completely broken by a `brew update`. That no-one has complained about them still being broken does make me pretty dubious about the value they are providing.

> After this, I will be able to finalize this PR so it can be ready to be merged.

Can this PR use the old Ruby until they are merged? I'd rather not have them merged until I can see that we have some sort of CI that can reliably build them.

If we have CI that demonstrates the ability to build them on my local machine using Docker: I'm begrudgingly OK with them being included. That's when @zbeekman's machine may be useful: building them on native hardware is going to be quicker for when we're uploading a new build.

I still question the value of providing these bottles when it appears their usage is low enough that in several weeks of people's installations being broken we've not had any bug reports/complaints.

sjackman:
> The problem with that approach is if you're bumping portable Ruby then anyone relying on the portable Ruby on ARM will have their Homebrew completely broken by a brew update.

We have until the next stable release to build the ARM bottles without breaking users' installation.

> That no-one has complained about them still being broken does make me pretty dubious about the value they are providing.

There's a chicken-and-egg problem here. To build a userbase for Homebrew on Linux on ARM, Homebrew on Linux needs to work on ARM. Homebrew on Linux on x86_64 took years to build a userbase (over 20k users now).

MikeMcQuaid:
> We have until the next stable release to build the ARM bottles without breaking users' installation.

In this case we missed that and have had two stable releases since.

> There's a chicken-and-egg problem here. To build a userbase for Homebrew on Linux on ARM, Homebrew on Linux needs to work on ARM. Homebrew on Linux on x86_64 took years to build a userbase (over 20k users now).

That's fair. I guess there's a wider debate there about "do we wish to build a user base for Homebrew on Linux on ARM" and, moreso, "does the lack of portable Ruby on ARM make it impossible to run Homebrew on Linux on ARM".

The Ruby bumps tend to coincide with macOS version bumps or security updates which are already fairly stressful times with time pressure. I'd really like it if we can find some way that allows people to run Homebrew on Linux on ARM that doesn't rely on this repository and Homebrew/brew having to provide portable rubies in a timely fashion. 

Perhaps `vendor-install` could source these URLs from e.g. a bash script in a tap which could also host the portable Ruby bottles? `vendor-install` could be updated to know about this script in this case and suggest that users run it or manually install their own Ruby if it's not available/updated yet.

xu-cheng:
All test passed.

sjackman:
>> We have until the next stable release to build the ARM bottles without breaking users' installation.

> In this case we missed that and have had two stable releases since.

With @xu-cheng 's infrastructure in place, the technical aspect of building these bottles should be easier now. Any of @xu-cheng @zbeekman and myself are able to provide the people power needed to update the bottles, so I believe we have the resources in place to make this update more timely next time around.

zbeekman:
That workflow is a thing of beauty. Nice work!

.github/workflows/tests.yml
```diff
@@ -4,40 +4,87 @@ on:
...
+      fail-fast: false
+      matrix:
+        arch:
+          - { name: x86_64, platform: linux/amd64, build_img: 'debian/eol:wheezy', test_img: 'ubuntu:xenial' }
+          - { name: arm32v6, platform: linux/arm/v6, build_img: 'resin/rpi-raspbian:wheezy', test_img: 'ubuntu:xenial' }
+          - { name: arm64v8, platform: linux/arm64/v8, build_img: 'debian/eol:jessie', test_img: 'ubuntu:xenial' }
+    steps:
+    - name: Set up Git repository
+      uses: actions/checkout@master
+    - name: Enable Docker experimental features
```
MikeMcQuaid: What are these for? Are they needed on x86_64? If not: can this be skipped on x86_64?

.github/workflows/tests.yml
```diff
@@ -4,40 +4,87 @@ on:
...
+  test-linux:
+    name: Build [Linux ${{ matrix.arch.name }}]
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - { name: x86_64, platform: linux/amd64, build_img: 'debian/eol:wheezy', test_img: 'ubuntu:xenial' }
+          - { name: arm32v6, platform: linux/arm/v6, build_img: 'resin/rpi-raspbian:wheezy', test_img: 'ubuntu:xenial' }
+          - { name: arm64v8, platform: linux/arm64/v8, build_img: 'debian/eol:jessie', test_img: 'ubuntu:xenial' }
```
MikeMcQuaid: if the `test_img` is always the same (or at least the same for all three): can it be hardcoded somewhere it's shared between all three?

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+RUN localedef -i en_US -f UTF-8 en_US.UTF-8 \
+    && useradd -m -s /bin/bash linuxbrew
+
+# hadolint ignore=DL3003
+RUN git clone --depth=1 https://github.com/Homebrew/brew /home/linuxbrew/.linuxbrew/Homebrew \
+    && (cd /home/linuxbrew/.linuxbrew/Homebrew; \
+        curl -L https://github.com/Homebrew/brew/pull/6619.patch -o /tmp/6619.patch; \
+        git config --global user.name BrewTestBot; \
+        git config --global user.email homebrew-test-bot@lists.sfconservancy.org; \
+        git am /tmp/6619.patch) \
```
MikeMcQuaid: Presumably this is the Homebrew/brew PR?

MikeMcQuaid:
> With @xu-cheng 's infrastructure in place, the technical aspect of building these bottles should be easier now. Any of @xu-cheng @zbeekman and myself are able to provide the people power needed to update the bottles, so I believe we have the resources in place to make this update more timely next time around.

Ok, I'll trust you all on that. Hopefully it's not for a while. Thanks all! A few comments and this should be good to go.

Dockerfile
xu-cheng: Yes, updated and removed.

.github/workflows/tests.yml
xu-cheng: This is needed to use the `--platform` flag. And it is required for all platforms.

MikeMcQuaid:
Great work here @xu-cheng!

sjackman:
Thank you for this work, @xu-cheng, and thanks for merging, Mike! 🎆

PR #95, opened 2020-02-05
tests: remove `brew tests`.

MikeMcQuaid:
These segfault on all supported platforms with Portable Ruby. I'm open to this being fixed but I've spent a bunch of time trying to do so in the past and it doesn't seem to bother non-maintainers.


MikeMcQuaid:
Merging despite ARM failures; they are broken on `master` and not by this PR.

PR #101, opened 2020-06-08
Bump minimum macOS requirement

mistydemeo:
This supports https://github.com/Homebrew/brew/pull/7698, where we're bumping the minimum OS version to Yosemite. I've updated the abstract formula and built a new macOS bottle, but we need a new Linux bottle.


Bo98:
I don’t mean to distract but if we’re revision bumping anyway, do we want to do a round of updates for deps like OpenSSL or nah? Not sure when that type of stuff is usually handled but I can go through and update some of them if desired.

mistydemeo:
That’s not a bad idea! If we can get that done pretty quickly then yes, I’ll rebuild this with those new deps.

MikeMcQuaid:
Yeh, I like the idea of bumping deps 👍. @Bo98 if you could do this your day today (either in this PR directly or another, up to you) then @mistydemeo when you're around if you could build and upload the versions for macOS and Linux x86_64 then I think we're good to go 🎉

PR #102, opened 2020-06-09
Updated dependencies

Bo98:
* portable-libyaml 0.2.2 -> 0.2.5
* portable-ncurses 6.1 -> 6.2
  - Also removed `dylib` symlinks which did nothing as we don't build dynamic libraries.
* portable-openssl 1.0.2t -> 1.1.1g
  - 1.1.x no longer compiles zlib by default for security reasons (CRIME vulnerability). I've switched the formula to respect that.
  - Also updated Curl cert bundle to 2020-01-01.
  - Added a pkg-config hack to make Ruby link with OpenSSL correctly. Ruby doesn't appear to properly support statically linking OpenSSL sadly, hence the need for the hack.
* portable-readline 8.0.1 -> 8.0.4
  - Removed redundant `SHLIB_LIBS` as it is for shared libraries.

### DIFF
--- a/Formula/portable-openssl.rb
+++ b/Formula/portable-openssl.rb
@@ -3,18 +3,15 @@ require File.expand_path("../Abstract/portable-formula", __dir__)
 class PortableOpenssl < PortableFormula
   desc "SSL/TLS cryptography library"
   homepage "https://openssl.org/"
-  url "https://www.openssl.org/source/openssl-1.0.2t.tar.gz"
-  mirror "https://dl.bintray.com/homebrew/mirror/openssl-1.0.2t.tar.gz"
-  mirror "https://www.mirrorservice.org/sites/ftp.openssl.org/source/openssl-1.0.2t.tar.gz"
-  sha256 "14cb464efe7ac6b54799b34456bd69558a749a4931ecfd9cf9f71d7881cac7bc"
-
-  depends_on "makedepend" => :build
-  depends_on "portable-zlib" => :build if OS.linux?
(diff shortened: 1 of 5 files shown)


Formula/portable-openssl.rb
```diff
@@ -47,57 +46,38 @@ def configure_args
...
-    # https://langui.sh/2015/11/27/sip-and-dlopen
-    if OS.mac?
-      inreplace "crypto/comp/c_zlib.c",
-                'zlib_dso = DSO_load(NULL, "z", NULL, 0);',
-                'zlib_dso = DSO_load(NULL, "/usr/lib/libz.dylib", NULL, DSO_FLAG_NO_NAME_TRANSLATION);'
-    end
-
     ENV.deparallelize
     system "perl", "./Configure", *(configure_args + arch_args)
-    system "make", "depend"
```
mistydemeo, line 79: Do we still need `makedepend` if this is gone?

Bo98: `make depend` still exists but is now implicitly done.

I'm not entirely if the `makedepend` dependency was needed in the first place. It's not used at all when building OpenSSL and Ruby in homebrew-core. I guess I can remove it and see if anything breaks.

MikeMcQuaid: Yeh, let's see what happens 👍

mistydemeo: Looks like it built fine!

MikeMcQuaid:
Great work here, thanks @Bo98!

PR #103, opened 2020-06-09
portable-ruby: copy ncurses terminfo into bottle

Bo98:
This is one step towards fixing `irb` on Linux not being able to handle left, right or backspace keys.


Formula/portable-ruby.rb
```diff
@@ -85,6 +85,8 @@ def install
       ENV["SSL_CERT_FILE"] ||= File.expand_path("../../libexec/cert.pem", RbConfig.ruby)
       #{openssl_rb_content}
     EOS
+
+    cp_r ncurses.opt_share/"terminfo", share/"terminfo" if OS.linux?
```
mistydemeo, line 89: Just checking if I understand: if this ended up misinstalled into `opt_share` somehow, we can probably do `(share/"terminfo").install ncurses.opt_share/"terminfo"` so we outright move it out of `opt_share`?

Bo98: It's not "misinstalled" per se. It's because we only bottle portable-ruby and not any of the dependencies.

What I'm doing here is simply copying files installed in portable-ncurses to also be in the portable-ruby Cellar.

Moving would have the same effect in the end (we don't care after this point if the portable-ncurses install still functions, since it will be discarded), but I don't think the sandbox will allow us to remove files from other formula install directories anyway. (Or if it does allow for some reason, I don't think it should do.)

MikeMcQuaid: I'd maybe do `nurses.share` instead of `opt_share` here.

Bo98: Yeah, fair point. I'll change the OpenSSL section too.

mistydemeo: Oh yeah, I'd totally misread this - thought you were copying from ncurses into ncurses, not into Ruby. 👍

Bo98:
Moving some of the discussion here from Slack:

Part two of fixing the issue involves setting `TERMINFO_DIRS` somewhere, and the question now is where - be it in `brew` or by patching Ruby. This cannot be baked in at compile time - it must be set at runtime, since it doesn't work well with relative paths.

MikeMcQuaid:
> Part two of fixing the issue involves setting `TERMINFO_DIRS` somewhere, and the question now is where - be it in `brew` or by patching Ruby. This value to set `TERMINFO_DIRS` to cannot be baked in at compile time - it must be evaluated at runtime, since it doesn't work well with relative paths.

The suggestion I liked the most from Slack adding `/usr/share/terminfo` to the search patch rather than bundling our of `terminfo` files. I don't feel strongly, though. I think we need to have something merged by e.g. 5pm Scotland time today so this isn't blocking the release of the next Homebrew version.

Bo98:
Just to summarise the different options here:

* Bundle `terminfo` and set `TERMINFO_DIRS` in `brew` to `<portable-ruby-dir>/share/terminfo`.
  - Can be done either globally in `brew.sh` or locally to `IRB` invokations (dev-cmd and debrew),
  - Is there any concern about this affecting subprocesses?
    - Probably not - especially if there's nothing interactive.
    - Otherwise, making `portable-ncurses` look for another env is a way to prevent that (or some sort of module detour to unset it when not needed)

* Bundle `terminfo` and set `TERMINFO_DIRS` in `portable-ruby`.
  - This requires hacking the Ruby source.
  - Where exactly?

* Bundle `terminfo` and patch ncurses to determine the data directory dynamically.
  - Likely bulletproof, but is ultimately still a patch.

* Don't bundle `terminfo` and pass a configure arg to ncurses to rely on a system installation.
  - The `brew` Docker image does not currently provide this.
  - Any concerns about version mismatches?
  - Any other locations on various distros beside `/usr/share/terminfo`?

PR #107, opened 2020-08-01
cmd/portable-package: fix for Homebrew.args changes 

Bo98:
Broke after Homebrew/brew#8144.

Also use `safe_popen_read`.


reitermarkus:
Thanks, @Bo98!

PR #115, opened 2021-01-18
portable-ruby: patch to set correct arch on Apple Silicon

dtrodrigues:
Without this patch, RUBY_PLATFORM returns just `-darwin20` instead of `arm64-darwin20`, which causes issues in `hardware.rb` in brew:

https://github.com/Homebrew/brew/blob/095798be405be47429877f9d92f416153f86e147/Library/Homebrew/hardware.rb#L87-L93
https://github.com/Homebrew/brew/blob/095798be405be47429877f9d92f416153f86e147/Library/Homebrew/hardware.rb#L109-L114

This is included within the generated `configure` script in ruby 2.7.2, but not in 2.7.1, and not in any 2.6.x releases.

### DIFF
--- a/Formula/portable-ruby.rb
+++ b/Formula/portable-ruby.rb
@@ -64,6 +64,10 @@ class PortableRuby < PortableFormula
 
     args << "--with-opt-dir=#{paths.join(":")}"
 
+    # see: https://github.com/ruby/ruby/pull/3272
+    # needed to set correct arch on Apple Silicon as part of RUBY_PLATFORM
+    inreplace "configure", "\"processor-name=powerpc64\"\n#endif",
+        "\"processor-name=powerpc64\"\n#endif\n#ifdef __arm64__\n\"processor-name=arm64\"\n#endif"
     system "./configure", *args
     system "make"
     system "make", "install"


mistydemeo (commented):
Double-checking, is this already upstreamed? Or is this something that would be fixed if `configure` were regenerated?

dtrodrigues:
It's already upstream and was backported to 2.7, but not 2.6. 

https://github.com/ruby/ruby/commit/f46929833923d96e4dc706945fd537a1caaebe66 is the backport, with the changes to `tool/m4/ruby_universal_arch.m4` and `configure.ac` being the relevant ones, so both of those would need to be patched prior to running `autoconf` again if we don't want to change `configure` directly.

MikeMcQuaid:
@dtrodrigues Nice work here!

> It's already upstream and was backported to 2.7, but not 2.6.

If the back port can be applied cleanly (i.e. you can use a GitHub commit as the `patch do`) then that'd be ideal. Otherwise, can you add the same information in an inline Ruby comment? Thanks!

dtrodrigues:
The configure.ac and other code was refactored significantly between 2.6 and 2.7 so the backport doesn't apply cleanly. The `inreplace` is effectively just https://github.com/ruby/ruby/pull/3272, so I'll update the comment to reflect that.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again @dtrodrigues!

PR #119, opened 2021-03-10
README: remove reference to fixed bug

Bo98:
Having heard no complaints for months, I think this was indeed fixed by #103 and Homebrew/brew#7718.


mistydemeo (approved):

Bo98:
This PR doubles up as a check to see if our CI is still working :~)

(It's not working)

Bo98:
Homebrew/brew#10821 will fix that.

PR #123, opened 2021-04-30
portable-openssl: update mirror

Bo98:
Homebrew/mirror will cease tomorrow.

### DIFF
--- a/Abstract/portable-formula.rb
+++ b/Abstract/portable-formula.rb
@@ -30,8 +30,8 @@ module PortableFormulaMixin
   end
 
   def test
-    assert_no_match(/Homebrew libraries/,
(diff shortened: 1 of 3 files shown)


MikeMcQuaid (approved):

Bo98:
Not sure what the preference is for dealing with DL3059. Do we even use layer caching? (We currently ignore DL3008 so I assume not.)

MikeMcQuaid:
@Bo98 Makes sense 👍🏻

Bo98:
Which bit makes sense? I'll just ignore DL3059 and someone else can make it their problem if they think it needs addressing :-)

MikeMcQuaid:
> I'll just ignore DL3059 and someone else can make it their problem if they think it needs addressing

Yup, thanks ❤️

PR #135, opened 2021-09-27
Dependency updates

Bo98:
Align some formulae with changes made in Homebrew/core (except where differences are intentional).

I've also disabled MD2 support in OpenSSL. This is a legacy protocol and we shouldn't be using this anyway, let alone explicitly enable it.


MikeMcQuaid (approved):
Looks great!

Bo98:
Added an extra bump of readline to 8.1.1.

Issue #118, opened 2021-03-10
Native extensions using Ruby C API fail to link with newer GCCs


Bo98:
Example: https://github.com/Homebrew/brew/issues/10784#issuecomment-794496570

This mostly prevents `brew prof --stackprof` from working, but does have a slight effect on other gems - such as `msgpack` and `racc` running in compatibility mode.

This only affects newer GCCs. It is an issue with GCC 9 on Ubuntu 20.04 but not GCC 5 on Ubuntu 16.04.

## Steps to reproduce

```
brew ruby -rmkmf -e "have_func('rb_require')"
```

(Note: this will create a `mkmf.log` in your current working directory)

## Expected result

```
checking for rb_require()... yes
```

## Actual result

On `homebrew/ubuntu20.04:latest` Docker container:

```
checking for rb_require()... no
```

## Relevant logs:

```
/usr/bin/ld: /home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.3_2/lib/libruby-static.a(vm_trace.o): relocation R_X86_64_32 against `.rodata' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: /home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.3_2/lib/libruby-static.a(array.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: /home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.3_2/lib/libruby-static.a(bignum.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: /home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.3_2/lib/libruby-static.a(class.o): relocation R_X86_64_32 against `.rodata.str1.8' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: /home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.3_2/lib/libruby-static.a(compar.o): relocation R_X86_64_32 against `.text' can not be used when making a PIE object; recompile with -fPIE
[...]

PR #136, opened 2021-09-27
portable-ruby 2.6.8

Bo98:
macOS Monterey beta 7 ships with Ruby 2.6.8 (on arm64 at least).

It's perhaps a tad early since it only landed with the last beta but I've created this PR to prepare to move to that.

In any case, I'd really like to address #118 before we build any bottles are built. I feel that issue will start becoming more apparent when we consider moving to newer versions of Ubuntu or if we happen to introduce a gem dependency that is more likely to trigger an issue (we're lucky it doesn't affect anything in CI at the moment).

### DIFF
--- a/Formula/portable-ruby.rb
+++ b/Formula/portable-ruby.rb
@@ -3,11 +3,10 @@ require File.expand_path("../Abstract/portable-formula", __dir__)
 class PortableRuby < PortableFormula
   desc "Powerful, clean, object-oriented scripting language"
   homepage "https://www.ruby-lang.org/"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Thanks for this!

MikeMcQuaid:
> In any case, I'd really like to address #118 before any bottles are built.

I'm fine with that as long as it beats the public release of Monterey. If it doesn't: we should build these bottles anyway.

Bo98:
Yep that’s fine. I’ll look into it a bit this week maybe.

PR #149, opened 2022-04-16
Add and use a portable libxcrypt

### LINKED ISSUE #147: Compatibility with distros without libcrypt.so.1
https://github.com/Homebrew/brew/issues/12880

It appears that our Portable Ruby build might not work with Arch Linux without libcrypt.so.1. We should handle this scenario.

Bo98:
This is to fix #147.


Bo98:
> lib/util-xbzero.c: In function ‘_crypt_explicit_bzero’:
> lib/util-xbzero.c:58:1: error: invalid argument to gimple call
> s
> memset (s, 0, len);
>
> lib/util-xbzero.c:58: confused by earlier errors

Me too.

Look like a compiler crash?

Bo98:
I cannot reproduce with 4.7.2 on Godbolt, so this seems specific to Wheezy's build of GCC.

Doesn't happen with `gcc-4.9-backport` and `gcc-mozilla`.

Bo98:
Hacky but it seems to work.

Dockerfile
```diff
@@ -11,6 +11,13 @@ ENV HOMEBREW_NO_AUTO_UPDATE=1
 # hadolint ignore=DL3002
 USER root
 
+RUN if [ "x${img}" = "xhomebrew/debian7" ]; then \
+      apt-get update \
+      && apt-get install -y --no-install-recommends gcc-4.9-backport=4.9.2-10+deb7u1 \
```
MikeMcQuaid, line 16: Thought: maybe we should do this on our global wheezy image?

Formula/portable-libxcrypt.rb
```diff
@@ -0,0 +1,52 @@
...
+
+class PortableLibxcrypt < PortableFormula
+  desc "Extended crypt library for descrypt, md5crypt, bcrypt, and others"
+  homepage "https://github.com/besser82/libxcrypt"
+  url "https://github.com/besser82/libxcrypt/releases/download/v4.4.28/libxcrypt-4.4.28.tar.xz"
+  sha256 "9e936811f9fad11dbca33ca19bd97c55c52eb3ca15901f27ade046cc79e69e87"
+  license "LGPL-2.1-or-later"
+
+  def install
+    ENV["HOMEBREW_CC"] = "/usr/lib/gcc-4.9-backport/bin/gcc" if ENV["HOMEBREW_ON_DEBIAN7"].present?
```
MikeMcQuaid, line 11: Can this get a comment explaining what's going on and why? It's a bit hacky, I agree, but seems fine for something statically linked until we migrate away from Wheezy.

MikeMcQuaid (approved):

Dockerfile
Bo98: Could do, though I think it's also a good idea to scope this to only what needs it, which is only portable-libxcrypt.

MikeMcQuaid: cool 👍🏻

PR #153, opened 2022-06-02
Add release workflow

Bo98:
A fully automated release workflow that builds and uploads Portable Ruby bottles from a workflow dispatch event.

The _only_ secrets of our own involved are the BrewTestBot GPG signing keys. Pushing, uploading etc all use the auto-generated GITHUB_TOKEN.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,111 @@
+name: Release
+
+on:
+  workflow_dispatch:
+
+env:
+  HOMEBREW_DEVELOPER: 1
+  HOMEBREW_NO_AUTO_UPDATE: 1
+
+permissions:
+  contents: write
+  packages: write
+
+jobs:
+  determine-tag:
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.determine-tag.outputs.tag }}
+    steps:
+      - name: Set up Homebrew
+        uses: Homebrew/actions/setup-homebrew@master
+        with:
+          test-bot: false
+
+      - name: Determine tag
+        id: determine-tag
+        run: |
+          brew ruby -e "$(cat<<EOS
+          require "formula"
+
+          tag = Formula["portable-ruby"].pkg_version.to_s
+
+          existing_tag = Utils.safe_popen_read("git", "tag", "--list", tag)
+          raise "Tag #{tag} already exists" if existing_tag.present?
+
+          puts "Preparing for #{tag} release."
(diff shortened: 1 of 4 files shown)


Bo98:
I have tested this (except the GPG part) on my fork and it seems to work.

.github/workflows/release.yml
```diff
@@ -0,0 +1,111 @@
...
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          gh release create "${TAG}" --notes "Portable Ruby ${TAG}" --target "$(git rev-parse --verify HEAD)"
+
+      - name: Upload to GitHub Releases
+        working-directory: bottles
+        env:
+          HOMEBREW_GITHUB_API_TOKEN: ${{ github.token }}
+        run: brew pr-upload --debug --upload-only --root-url=https://github.com/${GITHUB_REPOSITORY}/releases/download/${TAG}
```
MikeMcQuaid, line 108: Completely overkill but: might be nice if this could create the Homebrew/brew PR or output some instructions on what needs to be entered and in what file?

MikeMcQuaid (approved):
Great work here. Good to merge as-is, just have a few questions I'm interested in exploring post-merge.

.github/workflows/tests.yml
```diff
@@ -18,69 +18,19 @@ jobs:
...
-            docker run --rm -v $(pwd):/data -e HOMEBREW_RUBY_PATH=/data/portable-ruby/bin/ruby \
-              homebrew-portable-test /bin/bash -c "brew config"
-          fi
-
-      - name: Post cleanup
-        if: always()
-        run: |
-          brew test-bot --only-cleanup-after
-          rm -rvf bottles portable-ruby
+  # This job is used as a required status check, instead of requiring each build matrix element.
```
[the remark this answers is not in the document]
MikeMcQuaid: > It's not quite an issue yet but when 10.11-cross is ephemeral then the name of it will be different every time so it'll be impossible to make it a required status check.

Gotcha, could the comment clarify this? Thanks ❤️

.github/workflows/release.yml
MikeMcQuaid: > Some sort of brew automation would be nice, but I'd be more comfortable to first change things around on that end to be easier to machine-replace (and ideally in a format that'll never really change for style reasons etc).

👍🏻 agreed.

MikeMcQuaid (approved):

PR #155, opened 2022-06-03
Remove bottle block

Bo98:
It's not worth trying to manage bottle commits (which means turning off CI checks etc.) in the release workflow for what zero users are going to use.

This also reduces the number of secrets required to zero.


MikeMcQuaid (approved):

Bo98:
Removed all the extra rebuild-handling code too, since we don't do rebuilds ever (and if we do: our tooling needs fixed to support that since it mostly was picking the wrong rebuild before anyway).

PR #164, opened 2022-10-25
portable-ruby: restore compatibility for older Ubuntu

Bo98:
This regressed when updating CI to Ubuntu 22.04 as it's pulling in paths which only exist on 22.04.


iMichka (approved):

mistydemeo (approved):

MikeMcQuaid (approved):

iMichka:
Rerunning jobs to see why CI broke

Formula/portable-ruby.rb
```diff
@@ -7,6 +7,7 @@ class PortableRuby < PortableFormula
   url "https://cache.ruby-lang.org/pub/ruby/2.6/ruby-2.6.10.tar.xz"
   sha256 "5fd8ded51321b88fdc9c1b4b0eb1b951d2eddbc293865da0151612c2e814c1f2"
   license "Ruby"
+  revision 1
```
MikeMcQuaid, line 10: Are we gonna distribute this?

MikeMcQuaid (commented):
Good catch!

PR #169, opened 2023-03-30
Rename SSL_CERT_FILE to avoid overriding system SSL cert file

Bo98:
We set a default `SSL_CERT_FILE` because OpenSSL contains fixed paths baked in that cannot be avoided.

However setting this has the unintended effect of affecting _all_ OpenSSL usages (e.g. in curl) rather than just the one linked into our portable Ruby.

We need to set a `SSL_CERT_FILE` otherwise `open-uri` etc will break, so let's just rename the portable-openssl variable to something different.

https://github.com/Homebrew/brew/issues/14828

### DIFF
--- a/Formula/portable-openssl.rb
+++ b/Formula/portable-openssl.rb
@@ -46,6 +46,11 @@ class PortableOpenssl < PortableFormula
   end
 
   def install
+    # OpenSSL is not fully portable and certificate paths are backed into the library.
+    # We therefore need to set the certificate path at runtime via an environment variable.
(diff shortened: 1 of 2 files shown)


Bo98:
Ah adding stackprof to brew's Gemfile broke CI here. Probably should separate it into a gem group since it really doesn't need to be installed.

MikeMcQuaid:
> Probably should separate it into a gem group since it really doesn't need to be installed for 99% of use cases.

👍🏻 fine with me.

PR #176, opened 2023-08-21
workflows/release: use new GITHUB_OUTPUT file

Bo98:
`set-output` is deprecated


Bo98:
CI failure is a known brew bug.

PR #180, opened 2023-10-26
workflows: various cleanups

Bo98:
* Remove CI on `master` push. This is a low volume repository so spending an extra 20-25 minutes is a waste of self-hosted resources when we don't allow direct pushes without a pull request. If there's a concern of two simultaneous PRs causing problems only when merged together, we can enable "Require branches to be up to date before merging", which will solve those concerns with minimal annoyance given the PR volume.
* Remove run attempt from ephemeral names as we don't need that anymore.
* Scope permissions slightly more tightly
* Minor cleanup of unused stuff

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,14 +9,13 @@ permissions:
 env:
   HOMEBREW_DEVELOPER: 1
   HOMEBREW_NO_AUTO_UPDATE: 1
-  HOMEBREW_CHANGE_ARCH_TO_ARM: 1
 
 jobs:
   build:
     strategy:
       matrix:
         include:
-          - os: '10.11-cross-${{github.run_id}}-${{github.run_attempt}}'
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
> * Remove CI on `master` push. This is a low volume repository so spending an extra 20-25 minutes is a waste of self-hosted resources when we don't allow direct pushes without a pull request. If there's a concern of two simultaneous PRs causing problems only when merged together, we can enable "Require branches to be up to date before merging", which will solve those concerns with minimal annoyance given the PR volume.

Could we keep `brew test-bot --only-tap-syntax` running on a Linux runner for the `master` branch, please?

MikeMcQuaid:
Otherwise: all agreed, nice work 👍🏻

Bo98:
Done in https://github.com/Homebrew/homebrew-portable-ruby/commit/0f7b1d15ef6bec549017612f442ef410664698d4

PR #184, opened 2024-02-28
workflows/build: use new 11-arm64-cross runner

Bo98:
This sees the final retirement of the non-ephemeral macOS CI.

I've also got a completely recreated 10.11-cross image that is picked up automatically here without any workflow changes but is being tested for the first time with the CI runs in this PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
       matrix:
         include:
           - os: '10.11-cross-${{github.run_id}}'
-          - os: '11-arm64'
+          - os: '11-arm64-cross-${{github.run_id}}'
           - os: 'ubuntu-latest'
             container:
               image: ghcr.io/homebrew/ubuntu22.04:master


Bo98:
Will do some tests in 10.11 VM before merging but I don't anticipate any issues

MikeMcQuaid (approved):
Woohoo, nice work here @Bo98!

Bo98:
10.11 WindowServer under QEMU keeps crashing for me so will test it fully later before I make a release.

I think I accidentally used CLT 14.2 (which claims 10.13 support) when I intended to use 13.4 (claims 10.9 support), but I'll either adjust that or just ignore it given the linker has inserted 10.11 minimum into the binary fine so must surely work ok.

Bo98:
Thanks. Can you test the binary from CI too? https://github.com/Homebrew/homebrew-portable-ruby/actions/runs/8088378084

EricFromCanada:
Ran `brew reinstall -v --force-bottle portable-ruby--3.1.4.el_capitan.bottle.tar.gz` and then `brew test portable-ruby`. Hopefully that covers it.

Bo98:
Should do yeah. Thank you!

PR #192, opened 2024-05-15
portable-libyaml: move to GitHub tag

MikeMcQuaid:
This better matches the formulae in homebrew/core.

Inspired by https://github.com/Homebrew/homebrew-portable-ruby/pull/190


SMillerDev (approved):

Bo98:
This does make the 10.11 build slower. Honestly not sure why homebrew-core uses the git tag - seems like a mistake unless you found the reason.

MikeMcQuaid:
@Bo98 Ok, thanks, reverting.

Issue #203, opened 2024-05-28
Add livecheck and autobump workflows


MikeMcQuaid:
We're now going to start making Portable Ruby be the newest Ruby version(s) not ending in `.0`.

For example, we are now on Portable Ruby 3.3.1. We should upgrade to 3.3.2 when it is released.

When 3.4.0 is released, we should not upgrade to it. When 3.4.1 is released: we should.

This would ideally be done with `livecheck` blocks for all the `portable-*` formulae in this repository and an autobump workflow that opens PRs for new `portable-*` versions.

We will only make a new release/upload when we have new Portable Ruby versions (bar critical security vulnerabilities).

CC @Homebrew/core @nandahkrishna @samford @chenrui333 @SMillerDev who may all be interested in helping out here.

samford:
Just to confirm before creating PR(s) for `livecheck` blocks:

* Is it possible to implement the `portable-ruby` version restriction in the autobump workflow or should it be handled in the `livecheck` block? I tend to prefer for `livecheck` to return the latest version and for tools that use that information to handle version restriction but it's maybe less of an issue since this isn't core/cask.
* The aforementioned version restriction only applies to `portable-ruby` and we simply use the latest version for the other formulae, correct?

bevanjkay:
> Is it possible to implement the portable-ruby version restriction in the autobump workflow or should it be handled in the livecheck block?

The `throttle` dsl should be able to handle this if it's set to 2?

MikeMcQuaid:
> * Is it possible to implement the `portable-ruby` version restriction in the autobump workflow or should it be handled in the `livecheck` block? I tend to prefer for `livecheck` to return the latest version and for tools that use the information to handle version restriction but it's maybe less of an issue since this isn't core/cask.

I don't feel strongly either way.

> * The aforementioned version restriction only applies to `portable-ruby` and we simply use the latest version for the other formulae, correct?

Correct, the others should always be latest.

> The `throttle` dsl should be able to handle this if it's set to 2?

Stupid question: would this not mean that we wouldn't get an upgrade to e.g. 3.3.2?

bevanjkay:
It's my understanding that `throttle` takes the final part of the `version` when doing the comparison.
[...]

PR #204, opened 2024-05-28
Add livecheck blocks to formulae

samford:
As discussed in https://github.com/Homebrew/homebrew-portable-ruby/issues/203, this adds `livecheck` blocks to the formulae in this tap. The `portable-ruby` `livecheck` block returns the newest stable version, assuming that the version restriction will be handled in any associated workflow. If we end up needing to handle the version restriction in the `livecheck` block, I can update this or create a follow-up PR.

One other thing worth noting is that the `livecheck` block for `portable-openssl` is checking the latest release on GitHub whereas the formula in homebrew/core checks https://www.openssl.org/source/. Should we align the check with the core formula(e)?

### DIFF
--- a/Formula/portable-libffi.rb
+++ b/Formula/portable-libffi.rb
@@ -7,6 +7,11 @@ class PortableLibffi < PortableFormula
   sha256 "b0dea9df23c863a7a50e825440f3ebffabd65df1497108e5d437747843895a4e"
   license "MIT"
 
+  livecheck do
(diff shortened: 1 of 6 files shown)


MikeMcQuaid:
> Should we align the check with the core formula(e)?

Yes, please!

MikeMcQuaid (approved):
Thank again @samford!

PR #205, opened 2024-05-28
feat: add autobump flow

chenrui333:
Similar workflow as we used in core and cask, set check frequency to daily

relates to #203

### DIFF
--- a/.github/autobump.txt
+++ b/.github/autobump.txt
@@ -1,0 +1,6 @@
+portable-libffi
+portable-libxcrypt
+portable-libyaml
+portable-openssl
+portable-ruby
+portable-zlib
--- a/.github/autobump.txt
+++ b/.github/autobump.txt
@@ -1,6 +1,0 @@
-portable-libffi
-portable-libxcrypt
-portable-libyaml
-portable-openssl
-portable-ruby
-portable-zlib
--- a/.github/workflows/autobump.yml
+++ b/.github/workflows/autobump.yml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
     inputs:
       formulae:
-        description: Custom list of formulae to livecheck and bump if outdated
+        description: Custom list of portable-ruby formulae to livecheck and bump if outdated
         required: false
   schedule:
     # Daily at midnight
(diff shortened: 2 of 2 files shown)


MikeMcQuaid (approved):
Looks good, thanks @chenrui333 and @SMillerDev!

SMillerDev (approved):

.github/workflows/autobump.yml
```diff
@@ -0,0 +1,55 @@
...
+          test-bot: false
+
+      - name: Configure Git user
+        uses: Homebrew/actions/git-user-config@master
+        with:
+          username: ${{ (github.event_name == 'workflow_dispatch' && github.actor) || 'BrewTestBot' }}
+
+      - name: Get list of autobump portable-ruby formulae
+        id: autobump
+        run: echo "autobump_list=$(xargs < "$(brew --repo homebrew/homebrew-portable-ruby)"/.github/autobump.txt)" >> "$GITHUB_OUTPUT"
```
Bo98, line 46: Rather not have to duplicate names into an autobump.txt:

```suggestion
        run: echo "autobump_list=$(brew tap-info --json "${GITHUB_REPOSITORY}" | jq -r '.[].formula_names[]')" >> "$GITHUB_OUTPUT"
```

.github/workflows/autobump.yml
```diff
@@ -0,0 +1,55 @@
...
+
+env:
+  HOMEBREW_FORCE_HOMEBREW_ON_LINUX: 1
+
+permissions:
+  contents: read
+
+jobs:
+  autobump:
+    if: github.repository == 'Homebrew/homebrew-portable-ruby'
```
Bo98, line 26:
```suggestion
    if: github.repository_owner == 'Homebrew'
```

Minor but more concise

chenrui333: yeah, works for me 👍

.github/workflows/autobump.yml
chenrui333: adopted 👍

.github/autobump.txt
Bo98, line 1: Should be able to remove this file now

MikeMcQuaid (approved):
Looks good, thanks again @chenrui333 and everyone else for the help!

chenrui333:
need a rerun for the ci? not sure why it failed

PR #207, opened 2024-05-31
workflows/autobump: fix list collection

Bo98:
`jq` by default prints newlines but we were using the single-line `GITHUB_OUTPUT` syntax


MikeMcQuaid (approved):

MikeMcQuaid:
Merging this to see if it fixes the failing autobump workflow.

PR #212, opened 2024-06-04
Build fixes for ARM

Bo98:
Thanks to Apple Silicon, testing whether a Portable Ruby builds in a Linux ARM container takes just a few minutes.

* Scope glibc@2.13 reqiurement to Linux only
  - This effectively means that ARM support will require glibc 2.35 as a minimum, which seems reasonable to start clean for a new architecture and worry about our glibc formulae later down the line.
* portable-openssl: disable afalgeng
  - The test for it seems to fail in aarch64 containers. It might not necessarily will have in a theoretical future CI but it's not a feature we ever use so I opted to disable the feature entirely (Ruby's openssl gem does not allow engines to be enabled when using OpenSSL >=3.0).
  - All engines in general are deprecated in OpenSSL itself anyway afaik. Given the Ruby gem dropped support already, maybe we can slim the build further and just do `no-engine`?

### DIFF
--- a/Abstract/portable-formula.rb
+++ b/Abstract/portable-formula.rb
@@ -82,8 +82,12 @@ class PortableFormula < Formula
       keg_only "portable formulae are keg-only"
 
       on_linux do
-        depends_on "glibc@2.13" => :build
-        depends_on "linux-headers@4.4" => :build
+        on_intel do
+          depends_on "glibc@2.13" => :build
+          depends_on "linux-headers@4.4" => :build
+        end
+        # When we move from Ubuntu 22.04, on ARM we should add a dependency to glibc@2.35 and linux-headers@5.15.
+        # When doing so, remember to update the C++ Intel conditional in the portable-ruby formula.
       end
 
       prepend PortableFormulaMixin
--- a/Formula/portable-openssl.rb
+++ b/Formula/portable-openssl.rb
@@ -73,6 +73,7 @@ class PortableOpenssl < PortableFormula
       no-legacy
       no-module
       no-shared
+      no-engine
     ]
   end
 
--- a/Formula/portable-ruby.rb
+++ b/Formula/portable-ruby.rb
@@ -97,7 +97,7 @@ class PortableRuby < PortableFormula
         # Change e.g. `CONFIG["AR"] = "gcc-ar-11"` to `CONFIG["AR"] = "ar"`
         s.gsub!(/(CONFIG\[".+"\] = )"gcc-(.*)-\d+"/, '\\1"\\2"')
         # C++ compiler might have been disabled because we break it with glibc@2.13 builds
-        s.sub!(/(CONFIG\["CXX"\] = )"false"/, '\\1"c++"')
+        s.sub!(/(CONFIG\["CXX"\] = )"false"/, '\\1"c++"') if Hardware::CPU.intel?
       end
     end
 


Bo98:
> * All engines in general are deprecated in OpenSSL itself anyway afaik. Given the Ruby gem dropped support already, maybe we can slim the build further and just do `no-engine`?

Ended up just doing this. OpenSSL is probably the slowest to build so saves some time there.

SMillerDev (approved):

ZhongRuoyu (approved):
Thanks! A few notes from my experience playing around with https://github.com/ZhongRuoyu/homebrew-portable-ruby-aarch64-linux:

- > This effectively means that ARM support will require glibc 2.35 as a minimum, which seems reasonable to start clean for a new architecture and worry about our glibc formulae later down the line.

  If we want to have broader coverage, we may want to introduce `glibc@2.17`, the first Glibc with aarch64 support. Tested to work.
- Currently bottling things on aarch64 Linux can be painfully slow because `gnu-tar` is not bottled, and has a huge build dependency tree due to this dependency chain: `gnu-tar` -> `acl` -> `attr` -> `gettext` -> `libxml2` -> `python@3.11`, `python@3.12`, ...
- This may be handled by `brew`'s superenv, but if we care, GCC 9.3.1 and later enable out-of-line atomics by default, which may not have out-of-the-box support on earlier systems (https://learn.arm.com/learning-paths/servers-and-cloud-computing/lse/intro/).

MikeMcQuaid (approved):

MikeMcQuaid:
> * If we want to have broader coverage, we may want to introduce `glibc@2.17`, the first Glibc with aarch64 support. Tested to work.

@ZhongRuoyu This makes sense, can you submit a PR!

> * Currently bottling things on aarch64 Linux can be painfully slow because brewed `gnu-tar` is required but not bottled, and has a huge build dependency tree due to this dependency chain: `gnu-tar` -> `acl` -> `attr` -> `gettext` -> `libxml2` -> `python@3.11`, `python@3.12`, ...

We should just use the system `gnu-tar` on ARM instead. Similarly, a Homebrew/brew PR would be appropriate I think.

MikeMcQuaid:
Thanks @Bo98!

Bo98:
> If we want to have broader coverage, we may want to introduce glibc@2.17, the first Glibc with aarch64 support. Tested to work.

You would not only need to add a `glibc@2.17` but also update regular `glibc` to support being bootstrapped on aarch64 since that will be auto installed for 2.17-2.34.

It's something we would need to do eventually in any case if we ever moved from 2.35 in the future so if someone is able to do that prior to us adding support for aarch64 here then 2.17 minimum should be fine.

Bo98:
> * Currently bottling things on aarch64 Linux can be painfully slow because brewed `gnu-tar` is required but not bottled, and has a huge build dependency tree due to this dependency chain: `gnu-tar` -> `acl` -> `attr` -> `gettext` -> `libxml2` -> `python@3.11`, `python@3.12`, ...

That sounds like a ridiculous dependency tree. I might have to take a look at that...

Bo98:
Opened some Homebrew/core PRs to fix that dependency tree.

Though we still really should split the Python bindings from libxml2 as that dependency tree isn't great and libxml2 is used in a lot of formulae.

PR #216, opened 2024-06-12
Fix style issues

samford:
While fixing `brew style` tap support (https://github.com/Homebrew/brew/pull/17482), brew CI surfaced some style issues in this tap. This fixes the issues, so brew CI will pass and the fix can be merged.


MikeMcQuaid:
Thanks @samford!

PR #219, opened 2024-06-13
portable-ruby: slim bundled gems

Bo98:
* Remove all bundled gems that we are not using (i.e. everything except `debug`)
* Add a helper file so that we can replace this manual piece of code in Homebrew/brew: https://github.com/Homebrew/brew/blob/c997fa5327571db05bfcf5bd8a54370cf7a8e787/Library/Homebrew/standalone/init.rb#L43-L47 with a `require "portable_ruby_gems"`, which allows bundled gems to used

The functionality introduced here is mostly a nice-to-have for now but it may become very much central to our needs over time. Fiddle will be moved to be a bundled gem in Ruby 3.5.0 which is a gem we need on macOS so this will help with that as normally bundled gems cannot be used in Bundler environments and Fiddle also cannot be vendored in Homebrew/brew due to native extensions.

Relatedly, this system could allow us to include additional native gems should we need them. I could demonstrate that with Bootsnap if we want to test that out in the upcoming 3.3.3 release today. It should however never be used for anything critical on Linux as system Ruby is allowed there, so bundling things here should only be done for very few things (i.e. no more than what's mentioned in this PR). Fiddle is used on macOS only, `debug` is a niche feature that is already gated to be Portable Ruby only and Bootsnap auto-disables if not present.

### DIFF
--- a/Formula/portable-ruby.rb
+++ b/Formula/portable-ruby.rb
@@ -23,7 +23,47 @@ class PortableRuby < PortableFormula
     depends_on "portable-zlib" => :build
   end
 
+  resource "msgpack" do
+    url "https://rubygems.org/downloads/msgpack-1.7.2.gem"
+    sha256 "59ab62fd8a4d0dfbde45009f87eb6f158ab2628a7c48886b0256f175166baaa8"
+
+    livecheck do
+      url "https://rubygems.org/api/v1/versions/msgpack.json"
+      strategy :json do |json|
+        json.first["number"]
+      end
+    end
+  end
+
+  resource "bootsnap" do
+    url "https://rubygems.org/downloads/bootsnap-1.18.3.gem"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Makes sense to me! Good once CI is 🟢.

On bootsnap: I don't have strong feelings either way. It seems like it updates infrequently enough to make it worthwhile. I suspect if we ever want to enable it by default (which we should consider) we'll need to add better error handling around when it blows up (which I see periodically).

Bo98:
> I suspect if we ever want to enable it by default (which we should consider) we'll need to add better error handling around when it blows up (which I see periodically).

For sure. Before enabling we need to home in on what causes the breakage and potentially key the cache with that variable.

Long term ambition is to enable it by default, but that can only happen when we are comfortable it is stable for all.

PR #228, opened 2024-07-14
cmd/portable-package: add missing requires.

MikeMcQuaid:
These were no longer required by default in Homebrew/brew for speed.


MikeMcQuaid:
Needs rerun after https://github.com/Homebrew/brew/pull/17739 is merged.

PR #230, opened 2024-08-02
portable-ruby: patch to fix rare hang on forking

Bo98:
Under certain circumstances Ruby can hang in a `fork`ed child process with `__psynch_mutexwait` on the ractor scheduler lock.

This is very rare, but seemingly more common on macOS 14 x86_64 for unknown reasons.

Stripe have experienced the same issue on x86_64 Linux and have developed a patch that they are internally using. The patch is public on the Sorbet repository.

The bug appears to be a regression in Ruby 3.3.

### DIFF
--- a/Formula/portable-ruby.rb
+++ b/Formula/portable-ruby.rb
@@ -6,6 +6,7 @@ class PortableRuby < PortableFormula
   url "https://cache.ruby-lang.org/pub/ruby/3.3/ruby-3.3.4.tar.gz"
   sha256 "fe6a30f97d54e029768f2ddf4923699c416cdbc3a6e96db3e2d5716c7db96a34"
   license "Ruby"
+  revision 1
 
   # This regex restricts matching to versions other than X.Y.0.
   livecheck do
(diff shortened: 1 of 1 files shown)


p-linnane (approved):

carlocab (approved):

Formula/portable-ruby.rb
```diff
@@ -47,6 +48,13 @@ class PortableRuby < PortableFormula
     end
   end
 
+  # Fix rare hang on forking. Affected Homebrew/core CI, especially on macOS 14 x86_64.
+  # Patch in production use at Stripe.
```
MikeMcQuaid, line 52: Could this get link(s) to PRs/issues? thanks!

MikeMcQuaid (commented):
Great work @Bo98!

Formula/portable-ruby.rb
p-linnane: https://github.com/Homebrew/homebrew-portable-ruby/pull/231

PR #352, opened 2025-09-22
workflows/build: use 10.15 for x86_64 Ruby.

MikeMcQuaid:
We're dropping support for 10.11-10.14 soon so let's try this out.


p-linnane (approved):
Thanks!

PR #353, opened 2025-09-22
portable-formula: bump target to macOS Catalina

p-linnane:
We've dropped Mojave and earlier, so let's target Catalina.


MikeMcQuaid (approved):
Looks good!

